PR #4292, opened 2014-09-29
Validate publication dataset name, primary dataset name and request name. Fixes #4257.

AndresTanasijczuk:
This pull request is the same as #4277, which I had to "rebase" it. I copy/paste below the comment I made on that pull request.

The validity check that I added for General.requestName is maybe not needed, because it is already done in the server...
I just tried specifying a General.requestName with a white space and I get:

Will use configuration file crabConfig_test.py
WARNING: Incompatible CRABClient version "3.3.9"
Server is saying that compatible versions are: ['3.3.9.pre2']
Sending the request to the server
Error contacting the server.
Server answered with: Invalid input parameter
Reason is: Incorrect 'General.requestName' parameter
Log file is /afs/cern.ch/work/a/atanasi/CRAB3-tutorial/CMSSW_7_0_5/src/tests_vmatanasi2/crab_test_publishDataName_check_1 2/crab.log

The only improvement would be the message to the user, which is a bit clearer in the check I added.
Tell me @mmascher if you want me to remove the requestName validity check.

PR #4360, opened 2015-01-05
Add protection to avoid PrivateMC with inputDataset.

AndresTanasijczuk:
Also:
1) Require Data.primaryDataset in PrivateMC.
2) Don't accept Data.primaryDataset in Analisys job type.
3) Move the spellchecking of the configuration parameters from submit::\__call\__ to SubCommand::validateConfig (which if called in SubCommand::\__init\__ when loading the config)
4) Some small changes in messages to user when validating the config and in crab report.

### DIFF
--- a/src/python/CRABClient/Commands/SubCommand.py
+++ b/src/python/CRABClient/Commands/SubCommand.py
@@ -7,6 +7,7 @@ from ast import literal_eval
 
 from CRABClient.client_utilities import loadCache, getWorkArea, server_info, createWorkArea
 from CRABClient.client_utilities import BASEURL, SERVICE_INSTANCES
+from CRABClient import SpellChecker
 import CRABClient.Emulator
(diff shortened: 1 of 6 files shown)


AndresTanasijczuk:
I made the fix for using as default primaryDataset = "CRAB_PrivateMC" (in PrivateMC jobs) if publication is off. In that sense the pull request is now ready. 
But I think we can move already now the validation of the primaryDataset to the server, because the validation would not be done by matching a regular expression in the server, but instead by calling a function from WMCore. Therefore, we would have a new function in the server like we currently have for the outLFN parameter. What you think @mmascher ? The same is true for the other parameter I wanted to validate in the client, publishDataName. We can also already do it in the server, because we will use a function from WMCore. See https://github.com/dmwm/CRABClient/pull/4292/files

AndresTanasijczuk:
I removed the validation of the primary dataset name from here, because I will do it in the server. 
PR is now ready for merging.

PR #4373, opened 2015-01-18
Rename basicConfig to config.

AndresTanasijczuk:
This way the user can do

import CRABClient
config = CRABClient.Config()

as suggested by @belforte in https://github.com/dmwm/CRABServer/issues/4519. 
@belforte Now that we have introduced the module CRABClient.UserUtilities which contains all functions that we expose to users (including Config() -which is still called basicConfig()-), do you still prefer the above for the config, or you prefer to keep Config() inside CRABClient.UserUtilities?

Other question I have also is: If the user wants to use the functions from CRABClient.UserUtilities, how should we tell him/her to use these functions? Two options below:

1) Only import CRABClient and he/she gets everything, but then calling a function from CRABClient.UserUtilities requires to always put CRABClient.UserUtilities:

import CRABClient   <---- That's the only import the user will do.
config = CRABClient.Config()
hello = CRABClient.UserUtilities.func1()
ciao = CRABClient.UserUtilities.func2()

or

2) Use separate imports for the functions in CRABClient.UserUtilities:

import CRABClient
config = CRABClient.Config()
from CRABClient.UserUtilities import func1(), func2()
hello = func1()
ciao = func2()

### DIFF
--- a/src/python/CRABClient/UserUtilities.py
+++ b/src/python/CRABClient/UserUtilities.py
@@ -17,6 +17,20 @@ from WMCore.Configuration import Configuration
 from CRABClient.ClientExceptions import ClientException, UsernameException, ProxyException
 
 
+def config():
+    """
+    Return a Configuration object containing all the sections that CRAB recognizes.
+    """
+    config = Configuration()
+    config.section_("General")
+    config.section_("JobType")
+    config.section_("Data")
+    config.section_("Site")
+    config.section_("User")
+    config.section_("Debug")
+    return config
+
+
 def getUsernameFromSiteDB():
     """
     Retrieve username from SiteDB by doing a query to
@@ -62,20 +76,6 @@ def getUsernameFromSiteDB():
     if username == 'null':
         username = None
     return username
-
-
-def basicConfig():
-    """
-    Return a Configuration object containing all the sections that CRAB recognizes.
-    """
-    config = Configuration()
-    config.section_("General")
-    config.section_("JobType")
-    config.section_("Data")
-    config.section_("Site")
-    config.section_("User")
-    config.section_("Debug")    
-    return config
 
 
 def getFileFromURL(url, filename = None):


AndresTanasijczuk:
@belforte Any thoughts about my questions above?

belforte:
@AndresTanasijczuk I am confused, you say CRABClient.UserUtilities includes Config, then show examples with CRABClient.Config
about what we ask/suggest users to write, is import CRABClient needed anyhow ?

belforte:
seems to me users have a few equivalent ways:
1)
import CRABClient
hello=CRABClient.UserUtilities.f1()

2)
from CRABClient import UserUtilities
hello=UserUtilities.f1()

3)
fromCRABClient import UserUtilities as CrabUtils
hello=CrabUtils.f1()

4)
from CRABClient.ClientUtilities import *
hello=f1()

those are the same in terms of our code, only differ in the example, and since it is a matter of taste...
depends on the use case, e.g. how many many statements like hello=CRABClient.UserUtilities.f1() I have to type, if one: fine and clear, if 20  I'd like a shorter name.

Or maybe I am missing the point ?

AndresTanasijczuk:
On Wed, Jan 21, 2015 at 11:30 AM, Stefano Belforte <notifications@github.com
> wrote:

> @AndresTanasijczuk <https://github.com/AndresTanasijczuk> I am confused,
> you say CRABClient.UserUtilities includes Config,
>

It includes it until now. If we merge this PR, it will be moved to
CRABClient.__init__.py so that a user will have to write CRABClient.Config
as shown in my examples (and as you suggested in
https://github.com/dmwm/CRABServer/issues/4519)

The point of my questions was: are you ok with having one user exposed
function, namely Config(), in CRABClient.__init__.py while keeping all
other user exposed functions in CRABClient.UserUtilities.py? (Because I
would naively think that the best is to keep all of them in one module.)
Since we don't want to move all user exposed functions
to CRABClient.__init__.py, the other option I see is to keep Config() also
in CRABClient.UserUtilities.py as it is now. And then you have to give up
with your suggestion in https://github.com/dmwm/CRABServer/issues/4519.

belforte:
I see. 
We should put all user functions in one only place.
Sorry if I sounded like suggesting otherwise.
I do not like multiple import hierarchy unless there's a good reason, but if CRABClient already has too many functions, or user's config can avoid importing it, then to only import CRABClient.UserUtilities will be fine.  My suggestion from  https://github.com/dmwm/CRABServer/issues/4519 is recovered with case 3) of my list above

AndresTanasijczuk:
Ok! Then we keep all user exposed functions in CRABClient.UserUtilities.
And still minor question: do you want to rename the function from basicConfig() (this is the name it has now) to Config() or config()?

belforte:
since it is the only config function in that module, I would remove basic.
Whether initial is lower case or upper case, depends on the chosen convention,
I do not have one to recommend, but hopefully there is some standard at least
in Crab3, if not in WMDM

AndresTanasijczuk:
In general we use lower case for the first letter. Other user exposed functions use this convention. For example: getUsernameFromSiteDB() and getFileFromURL().

AndresTanasijczuk:
Fixes https://github.com/dmwm/CRABServer/issues/4519.

PR #4376, opened 2015-01-20
bug in passing filecacheurl variable

juztas:
It does not work if you just add filecacheurl, we found that it requires to pass it like this

### DIFF
--- a/src/python/CRABClient/Commands/submit.py
+++ b/src/python/CRABClient/Commands/submit.py
@@ -103,7 +103,7 @@ class submit(SubCommand):
         crab_job_types = getJobTypes()
         if upper(configreq['jobtype']) in crab_job_types:
             plugjobtype = crab_job_types[upper(configreq['jobtype'])](*pluginParams)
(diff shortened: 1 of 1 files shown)


AndresTanasijczuk:
Given @mmascher change here
https://github.com/dmwm/CRABClient/blob/master/src/python/CRABClient/JobType/PrivateMC.py#L17-21
I suspect he also wanted (but forgot) to change
https://github.com/dmwm/CRABClient/blob/master/src/python/CRABClient/JobType/Analysis.py#L27
to 
def run(self, **kwargs):
and then add in (or before) line https://github.com/dmwm/CRABClient/blob/master/src/python/CRABClient/JobType/Analysis.py#L78
filecacheurl = kwargs.get('filecacheurl')

mmascher:
I am sorry guys, I forgot to commit https://github.com/dmwm/CRABClient/commit/8794cfd52a7fc96dad106c30222ff43ef945729b last night

PR #4382, opened 2015-01-23
Move check of renamed parameters from submit to SubCommand.

AndresTanasijczuk:
When I moved the spell checking to the SubCommand validateConfig(), this was taking precedence over the check of renamed parameters, which was done later in the validateConfig of the submit command.
I think it makes sense to move the check of renamed parameters to the SubCommand validateConfig, as it is a general config validation necessary not only for the submit command. And we should do it before the spell checking. I also moved the check of the sections definition to the SubCommand validateConfig.

### DIFF
--- a/src/python/CRABClient/Commands/SubCommand.py
+++ b/src/python/CRABClient/Commands/SubCommand.py
@@ -10,7 +10,7 @@ from CRABClient.ClientUtilities import BASEURL, SERVICE_INSTANCES
 from CRABClient import SpellChecker
 import CRABClient.Emulator
 from CRABClient.ClientExceptions import ConfigurationException, MissingOptionException, EnvironmentException, UnknownOptionException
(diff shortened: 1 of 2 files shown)


AndresTanasijczuk:
$ crab submit -c crabConfig_test.py
Will use configuration file crabConfig_test.py.
Warning: The method getBasicConfig is now available in the CRABClient.UserUtilities module (with new name 'config'). From 1 April 2015, loading getBasicConfig from CRABClient.client_utilities will not be possible anymore (CRABClient.client_utilities will be removed). Please change your code accordingly before 1 April 2015.
Invalid CRAB configuration: Parameter Data.outlfn has been renamed to Data.outLFN; please change your configuration file accordingly.
The documentation about the CRAB configuration file can be found in https://twiki.cern.ch/twiki/bin/view/CMSPublic/CRAB3ConfigurationFile.

PR #4395, opened 2015-02-09
Some changes in resubmit command.

AndresTanasijczuk:
Don't harcode list-parameters in resubmit command.


AndresTanasijczuk:
New things added:
- Interpret empty string in site white and black list options as an empty list.
- Rename options to have the same name as the server parameters. Makes it more uniform along the code.
- Add comment about what is the convention I used for overwriting parameters when resubmitting jobs.

PR #4399, opened 2015-02-13
Improve error messages in getUsernameFromSiteDB. Use subprocess.Popen instead of commands.getstatusoutput.

AndresTanasijczuk:


belforte:
I will let @ericvaandering preach on the bad habit of using eval to turn json into a dictionary, at least hoping sitedb returns json and not some comma separated key:value which looks a lot like a python dictionary. Anyhow I understand the question was not for me and surely one strip too many should not harm.

ericvaandering:
I was all set to do that on the comment earlier today, but "from ast import literal_eval" claims to be totally safe and only parses data structures. :-)

mmascher:
I mentioned you Stefano because I thought this code was "ported" from CRAB2 and the original author was you :)

belforte:
On 02/16/2015 09:28 PM, Marco Mascheroni wrote:
> I mentioned you Stefano because I thought this code was "ported" from CRAB2 and the original author
> was you :)

oh no.  Crab2 code uses curl|awk|cut|grep|tr|..., when it comes to this
kind of stuff I like to do something that can be reproduced command line.
Problem is always that this gets executed in the user environment, and for
as much as we can document etc... and say "it works on lxplus so it is your
fault", in the end we have to figure out what they are doing wrong, whether
it is their env., op.sys., hardware, shell....

belforte:
safe as "not explotable by hackers", yes. Safe as "this is really a python dictionary what I am 
trying to eval", well... usual question: is SiteDB returning JSON or output-of-print ? Fear is that
like DBS it returns a dictionary, not JSON. Fine, as long as it is clear.

PR #4401, opened 2015-02-16
Remove unnecessary string.strip(username).

AndresTanasijczuk:


mmascher:
It would be intresting to know why this strip was put there in the first place. @belforte ?

AndresTanasijczuk:
Notice that in https://github.com/dmwm/CRABClient/pull/4399 I changed the way we extract the username from the curl command output. Before we were doing pure bash: "curl ... | tr ':,' '\n' | grep -A1 login | tail -1" which was resulting in username = ' "atanasi"', and therefore we needed to strip the leading white space. Now I take the curl output and do literal_eval on it, which gives me a python dictionary from which I take dict[result][0]['login'] to extract the username (of course without white space). This is an example of the curl output:
{"result": [
 {"dn": "/C=DE/O=GermanGrid/OU=DESY/CN=Till Arndt", "login": "tarndt", "method": "X509Proxy", "roles": {}, "name": "Till Michael Arndt"}
]}
When doing "tr ':,' '\n'" on it, we end up with 2 following lines:
"login"
 "atanasi"   <--- there is a white space before "atanasi"

PR #4402, opened 2015-02-17
Change Data.userInputFile to Data.userInputFiles, and accept a python list instead of a text file.

nizamyusli:
Fixes https://github.com/dmwm/CRABServer/issues/4244

### DIFF
--- a/src/python/CRABClient/JobType/Analysis.py
+++ b/src/python/CRABClient/JobType/Analysis.py
@@ -4,6 +4,7 @@ CMSSW job type plug-in
 
 import os
 import tempfile
+import string
 import re
 
 from WMCore.DataStructs.LumiList import LumiList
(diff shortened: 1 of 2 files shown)


src/python/CRABClient/JobType/Analysis.py
```diff
@@ -92,14 +92,11 @@ def run(self, filecacheurl = None):
...
         # Upload list of user-defined input files to process as the primary input
-        userFileName = getattr(self.config.Data, 'userInputFile', None)
-        if userFileName:
-            self.logger.debug("Attaching a list of user-specified primary input files from %s." % userFileName)
-            fnames = []
-            for fname in open(userFileName).readlines():
-                fnames.append(fname.strip())
-            configArguments['userfiles'] = filter(lambda x: x, fnames) #removing whitelines and empty objects
-
+        userFilesList = getattr(self.config.Data, 'userInputFiles', None)
```
AndresTanasijczuk: Should we make a set of this list?

mmascher: It makes sense to me. Unless there are use cases where people want to run two jobs on the same input file... In that case we should just print a warning. but I can't see a use case for this, so let's make it a set and print a warning.

PR #4404, opened 2015-02-18
Add a 'dryrun' option for crab submit.

annawoodard:
Example output:

```
Sending the request to the server
Success: Your task has been delivered to the CRAB3 server.
Waiting for task to be processed
Checking task status
Task status: NEW
Please wait...
Task status: UPLOADED

Executing local test run (needed for timing estimates), please wait...
Using LumiBased splitting

Task consists of 301 jobs to process 3009 lumis
The estimated memory requirement is 1306 MB
The longest job will process 10 lumis, with an estimated processing time of 142 minutes
The average job will process 9 lumis, with an estimated processing time of 142 minutes
The shortest job will process 9 lumis, with an estimated processing time of 128 minutes

Dry run requested: task paused
To continue processing, use 'crab proceed'
```

### DIFF
--- a/src/python/CRABClient/ClientMapping.py
+++ b/src/python/CRABClient/ClientMapping.py
@@ -93,7 +93,8 @@ commandsConfiguration = {
     'resubmit'      : {'requiresREST': True,  'initializeProxy': True,  'requiresTaskOption': True,  'useCache': True },
     'status'        : {'requiresREST': True,  'initializeProxy': True,  'requiresTaskOption': True,  'useCache': True },
     'uploadlog'     : {'requiresREST': True,  'initializeProxy': True,  'requiresTaskOption': False, 'useCache': False},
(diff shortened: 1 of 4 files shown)


ericvaandering:
I'm sorry I haven't been following this. But this output looks really useful. What would be even better would be two things:

1) Do the math for the user and suggest the "correct" splitting to get to our recommended 8 hr jobs.
2) Much more difficult, but allow them to change the splitting in the config file and re-split the task (or I guess delete the current task and replace it transparently)

Anyhow, it looks really useful!

annawoodard:
Thanks, Eric! I completely agree. I'm opening these in a separate issue: #4406.

mmascher:
I think we are ready to go, there is still one thing that I would like to improve, but we can file another ticket for it: instead of parsing the dagman input file inside `getDryRunOpts`, I would like to take the information we need from  the task database and/or from some adhoc json file we upload from the TW.

PR #4409, opened 2015-02-23
Accept ranges and wildcards for compatible-version in external configuration (gist)

nizamyusli:
Fixes: https://github.com/dmwm/CRABServer/issues/4213
Ranges and wildcards are now possible.
Accepted wildcard example: "3.3.3\*". This will be intepret as 3.3.3, 3.3.3.pre2, 3.3.3.patch1 etc.
Accepted range example: "3.3.1-3.3.3". This will be intepret as 3.3.1\*, 3.3.2\*, 3.3.3
                                           "3.3.1-3.3.3\*".This will be intepret as 3.3.1\*, 3.3.2\*, 3.3.3\*
Msg for the user will still show the one from gist and not from the long list.

### DIFF
--- a/src/python/CRABClient/Commands/SubCommand.py
+++ b/src/python/CRABClient/Commands/SubCommand.py
@@ -1,5 +1,6 @@
 import os
-import imp
+import re
+import imp 
 import json
 import types
 from optparse import OptionParser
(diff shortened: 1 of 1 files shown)


nizamyusli:
Re write the code to accept regular expression

PR #4411, opened 2015-03-02
crab getlog --short

nizamyusli:
Fixes https://github.com/dmwm/CRABServer/issues/4637
User can do 'crab getlog --short -dir <task name> --jobids=<job ids>' to get the short version of the log file.


mmascher:
Merging, but let's not publicize it to the user yet. We need to solve the "firewall issue" first.

PR #4413, opened 2015-03-03
Show condor cluster.id in crab status --long

nizamyusli:
Fixes: https://github.com/dmwm/CRABServer/issues/4676


nizamyusli:
Edit: Show column cluster id only in log file.

nizamyusli:
Edit: add table jobid/cluster id pair in log file.

PR #4414, opened 2015-03-05
Reduce proxy interactions

nizamyusli:
Fixes: https://github.com/dmwm/CRABServer/issues/4068
Check proxy only once every ten minutes.

### DIFF
--- a/src/python/CRABClient/Commands/SubCommand.py
+++ b/src/python/CRABClient/Commands/SubCommand.py
@@ -1,6 +1,7 @@
 import os
 import re
-import imp 
+import imp
+import time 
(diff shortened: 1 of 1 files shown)


mmascher:
@belforte why are you assigned? Should we hand this over to @emaszs ?

belforte:
I think it is assigned to me in order to review. We should  close this as no good and go back to issue to address the concerns you put there or drop the item.

belforte:
see previous comment https://github.com/dmwm/CRABClient/pull/4414#issuecomment-221872034
this is not a good change, and would not solve any problem.

PR #4419, opened 2015-03-19
Give error if arguments are given but command doesn't take arguments.

AndresTanasijczuk:


AndresTanasijczuk:
You can wait with this one. I can separate issues in separate pull request.

AndresTanasijczuk:
That's fine. No error will be given. But if you do 'crab status taskdir bla', an error will be given.

AndresTanasijczuk:
Note: I reordered commandsConfiguration in ClientMapping so that commands appear alphabetically ordered, but old key values have not changed.

PR #4428, opened 2015-03-20
Move condor restarts to a separate column in 'crab status --long'. Fixes #4379

### LINKED ISSUE #4379: What should the column "Retries" be in the output of 'crab status --long'?
Currently the 'crab status --long' output has a column called "Retries" which shows the sum of crab retries + condor restarts (https://github.com/dmwm/CRABClient/blob/master/src/python/CRABClient/Commands/status.py#L318), but when we use the --sort option it is only the crab retries (https://github.com/dmwm/CRABClient/blob/master/src/python/CRABClient/Commands/status.py#L317).

AndresTanasijczuk:


bbockelm:
Hi,

I'm uncomfortable with this one - some sites can have systemic issues that it's important to make visible to users.  This way they can be blacklisted.

What about reporting both separately?

Brian

AndresTanasijczuk:
Yeah, I actually wanted to report them separately. 
I want the user to clearly see how many times his job was submitted/resubmitted and that the retries column matches what the postjob log shows at the beginning, where is says "This is retry number <DagRetry>. The maximum number of retries is <MaxRetries>".
And then, ideally what I would like to have is a report that shows the history of the retries (showing each retry ala dashboard) and for each retry you would have the number of restarts. But maybe that's too much..? I added the restarts in another column:

Job State        Most Recent Site        Runtime   Mem (MB)      CPU %    Retries   Restarts      Waste  Exit Code
   1 finished     T1_RU_JINR              0:01:12             11                10                1              0    0:00:40          0

PR #4429, opened 2015-03-21
Add proceed command and fix typo (dirk -> dir) in crab-bash-completion.sh

AndresTanasijczuk:


mmascher:
FYI: we have a script that automatize the generation of the "complete" script. I need to modify the specfiles of the RPM so that we call it and the commands are automatically generated.

PR #4433, opened 2015-03-24
Allow wildcards in site names in resubmit blacklist/whitelist options; in that case no regex check is done in the client.

AndresTanasijczuk:


mmascher:
We can probably drop the whole regular expression check there. It's is already done in the server IIRC.

AndresTanasijczuk:
Yes, but I thought you wanted to keep it, given the comment you put in the code:
        #Checking if the sites provided by the user are valid cmsnames. Doing this because with only the
        #server error handling we get:
        #    Server answered with: Invalid input parameter
        #    Reason is: Incorrect 'siteblacklist' parameter
        #which is not really user friendly.
        #Moreover, I prefer to be independent from Lexicon. I'll the regex here.

If you agree, then I can drop the regular expression check.

mmascher:
True, let's keep the check until we start using the new WMCore and rewrite the messages in the server (actually it should be possible now!)

Basically now you can pass custom error messages to the validate_str[num...] functions. Let's drop the check on May wehn hopefully we will rewrite custom messages in the server

Issue #4325, opened 2014-11-21
voGroup option has no effect


bkargoll:
In my crabConfig.py I have the following lines:  
    config.section_("User")
    config.User.voGroup = 'dcms'

This should correspond to the dcms group I specify when creating my proxy and worked fine in Crab2. Nonetheless, when calling any crab command, I always get a query about my VO group:  
    $ crab checkusername
    Proxy file /tmp/x509up_u36073 exists with VO group = 'dcms' and VO role = 'NULL', but you have specified to use VO group = '' and VO role = 'NULL'.
    Do you want to overwrite the proxy (Y/N)?

I can explicitly give
    --voGroup dcms
in the command line for each crab command, and then in works fine.

AndresTanasijczuk:
Closing this issue, as we have decided to keep doing the check on the proxy VO group/role, but only for some commands (submit, checkwrite, grtlog, getoutput) and this was done in #4434

Issue #4326, opened 2014-11-22
Don't ask for overwriting proxy if user doesn't specify voRole/voGroup


AndresTanasijczuk:
If the user doesn't specify a voGroup or voRole (in command line or config) that is different than in the existing proxy, then we shouldn't give the message 

Proxy file /tmp/x509up_u.... exists with VO group = 'hola' and VO role = 'ciao', but you have specified to use VO group = '' and VO role = 'NULL'.
Do you want to overwrite the proxy (Y/N)?

but accept the proxy as it is.

AndresTanasijczuk:
Ok, we can keep the check for the submit command.

The point was that it is annoying that a user with a proxy that has a given VO group has to always use the --voGroup option in a crab command, otherwise CRAB will tell:

Proxy file /tmp/x509up_u.... exists with VO group = 'dcms' and VO role = 'NULL', but you have specified to use VO group = '' and VO role = 'NULL'.

I made a pull request where we control which commands should do the proxy VO group/role check and which ones not. If this pull request is ok, we could close this issue as well as #4325

AndresTanasijczuk:
Closing this issue, as we have decided to keep doing the check on the proxy VO group/role, but only for some commands (submit, checkwrite, grtlog, getoutput) and this was done in https://github.com/dmwm/CRABClient/pull/4434

PR #4434, opened 2015-03-25
Keep track of where were the voGroup and voRole set (config, command line opts, request cache) to give more clear messages to users. Not all commands require --voGroup/--voRole options. Not all commands need to check proxies VO group/role.

AndresTanasijczuk:
@mmascher Please check in the ClientMapping if you want to turn on/off the proxy check for some other commands.


AndresTanasijczuk:
Fixes #4325, #4326

PR #4443, opened 2015-04-13
Add message in crab resubmit.

AndresTanasijczuk:


belforte:
I would remove "succesfully" from the first line, since it was gives the expectation that submission will work.
https://github.com/AndresTanasijczuk/CRABClient/blob/92cf61dd3e900f4403f10090f6d3f6e94f5f30d4/src/python/CRABClient/Commands/resubmit.py#L63
Expecially if the next message is the error one, it can be confusing.

AndresTanasijczuk:
Agree. Changed.

PR #4446, opened 2015-04-15
Allow to not collect EDM and TFileService automatically. Fixes #4298

### LINKED ISSUE #4298: Allow to not collect EDM and TFileService automatically
Currently EDM and TFileService output files are collected automatically. It is not uncommon that users produce outputs that they don't care about in the end. For example: https://hypernews.cern.ch/HyperNews/CMS/get/crabFeedback/7812.html
Talking about that with @belforte, he suggested we introduce a new parameter in the configuration file, which will be by default False and so the user has to take care of putting in the already existing config parameter "outputFiles" the list of output files he wants to collect. If he sets instead the new parameter to True, then we automatically collect EDM and TFileService outputs (as it is done now). Or maybe we can have two parameters, one for EDM files and one for TFileService outputs.

nizamyusli:


mmascher:
I think it's ok. I will merge it after today's meeting though, I still have a question.

annawoodard:
I would change ignoreAutomaticOutputCollection --> disableAutomaticOutputCollection

AndresTanasijczuk:
This pull also needs a patch in PostJob, which I am preparing and testing.

nizamyusli:
Rebase and edited.

PR #4447, opened 2015-04-15
Add computing-tools email to socket-error message. Fixes #4340

### LINKED ISSUE #4340: Don't expect users to know how to "notify AnaOps"
https://github.com/dmwm/CRABClient/blob/3.3.12.rc5/bin/crab#L183
We should give the email address or HyperNews forum name. And don't use technical abbreviations like "AnaOps".

AndresTanasijczuk:


mmascher:
Did you see this happening @AndresTanasijczuk ?

AndresTanasijczuk:
No, I didn't.

mmascher:
ok, I actually think this exception is never raised. It was when we were using httplib instead of pycurl, but it might be deprecated now.

Has anybody ever seen the message 'Could not establish a connection with the server. This may be a temporary problem, but if you see it repeatedly please notify AnaOps' ?

We can get rid of this piece of code

AndresTanasijczuk:
Closing this pull, will get rid of exception.

PR #4461, opened 2015-04-22
getlog/getoutput use new API to retrieve transfer flags from TaskDB. Fixes #4369

### LINKED ISSUE #4369: getlog/getoutput use new API to retrieve transfer flags from TaskDB
Needs https://github.com/dmwm/CRABClient/issues/4368 to be done first.

nizamyusli:


nizamyusli:
Maybe can suggest a better warning msg to the user if there is no tm_edm_outfiles, tm_tfile_outfiles, tm_outfiles in the database.

AndresTanasijczuk:
Maybe @belforte can suggest a good message to put in https://github.com/nizamyusli/CRABClient/commit/365a0845cd9c60efd671f34817e24a1ab0d6c70e#diff-1259e30c4a45cc5db4edceac9c55d399R60 This is for the case in which the user didn't specify to collect any output file and CRAB client didn't detect any output file automatically in the pset. So getoutput will of course not retrieve anything. We want to give a good warning message to the user. Unfortunately we don't know if the user had specified disableAutomaticOutputCollection = True or False at submission time (because this parameter is not saved in the Oracle DB). So the message has to be generic, to cover both cases. In the first case it is users responsibility to provide the output files names; in the second case CRAB will detect PoolOutputModule and TFileService output files, but not other type of output files, which the user has to then specify.

belforte:
A general note: putting all possible present and future parameters in Oracle is indeed a pain and possibly a waste, couldn't we save the user config in the UFC ?
About the specific:

1. couldn't/shouldn't we detect this situation (no know output file) already at submission time and warn the user ? A task with no output is quite odd, unless it is test like HC.

2. message could be
 "no output can be retrieved because Crab could not detect any in the CMSSW configuration nor was any explicitly specified by user in crab configuration"

If we want to go in more details, let's do in a FAQ.

AndresTanasijczuk:
Thanks @belforte  
About 1. yes we can give a warning message already at submission time, but the task should run anyway, even if we think is odd, right?

belforte:
Right 
On May 3, 2015 11:38 AM, Andres Tanasijczuk <notifications@github.com> wrote:Thanks @belforte
About 1. yes we can give a warning message already at submission time, but the task should run anyway, even if we think is odd, right?

PR #4472, opened 2015-04-30
CRAB is running over full dataset when lumimask & runrange = null. Fixes #4471.

### LINKED ISSUE #4471: CRAB is running over full dataset when lumimask & runrange = null
While it should not run over anything.

nizamyusli:


src/python/CRABClient/JobType/Analysis.py
```diff
@@ -148,7 +148,7 @@ def run(self, filecacheurl = None):
         run_ranges_is_valid = run_ranges is not None and isinstance(run_ranges, str) and re.match('^\d+((?!(-\d+-))(\,|\-)\d+)*$', run_ranges)
         if run_ranges_is_valid:
             run_list = getRunList(run_ranges)
-            if lumi_list:
+            if lumi_list != None:
                 lumi_list.selectRuns(run_list)
```
ericvaandering: Are you sure this works? I think the key is here. If lumi list and run range were both specified but have no overlap, then lumi_list should be empty. Maybe this works because {} (empty) is not None?

nizamyusli:
There still need to be a fix in the server.

src/python/CRABClient/JobType/Analysis.py
```diff
@@ -156,7 +156,7 @@ def run(self, filecacheurl = None):
                     msg += " When Data.lumiMask is not specified, Data.runRange can not include more than 50000 runs."
                     raise ConfigurationException(msg)
                 lumi_list = LumiList(runs = run_list)
-        if lumi_list:
+        if lumi_list != None:
```
AndresTanasijczuk: The key is in this line. If lumi_list is empty (but not None), we still want to pass to the server this information about empty lists of runs and lumis (currently we are not passing it, and the server runs over the full dataset). Or should the client already not submit to server?

ericvaandering: Well, it seems to me at this point the client knows nothing useful will get done. It shouldn't need to round trip to the server to say "Hey, there is no work to do!"

AndresTanasijczuk:
Actually, what should we do if the user gives an empty lumi-mask and/or run-range? I mean, if he explicitly writes:
config.Data.lumiMask = {} (not syntactically correct, but the idea)
config.Data.runRange = ''
Should this be accepted or not? If yes, should it be interpreted as "don't use any lumi-mask/run-range" or as "filter all runs/lumis out" = "nothing to run over"? I would say should be accepted and interpreted as "don't use any lumi-mask/run-range". As the defaults for these parameters.

PR #4475, opened 2015-05-03
Give warning message in case no output file was detected in the CMSSW pset nor was any specified in the CRAB configuration.

AndresTanasijczuk:
@belforte Do you want to suggest a better message?


mmascher:
Why was the "in review process" label set?

AndresTanasijczuk:
Because I asked @belforte if he wanted to suggest a better message. But I think the one I put is already good. Removing the label.

Issue #4418, opened 2015-03-11
crabCommand should be quiet


ericvaandering:
Writing my first multiCRAB program I see that all the normal output from crab status, etc is printed to stdout. It should, IMO, be quiet by default and if needed a flag can be added to get that output. I'm parsing the JSON results of the commands and want my own printouts unmolested.

AndresTanasijczuk:
As suggested by @mmascher I documented in https://twiki.cern.ch/twiki/bin/view/CMSPublic/CRABClientLibraryAPI#Multicrab_using_the_crabCommand that user can do getLoggers(60) if he want crabCommand to be quiet.
Should we close this issue?

AndresTanasijczuk:
@mmascher are you satisfied with the documentation? 
@ericvaandering are you ok with Marco's proposal?

ericvaandering:
GetLoggers(60) seems like some black magic incantation. 

I suspect "60" is some magic threshold that silences most (all) messages, but it seems to me a) the function name is misleading and b) I should be able to import a constant and set that so something more like

from .... import setLogLevel, noLogInfo

setLogLevel(noLogInfo)

AndresTanasijczuk:
Ok @ericvaandering 
This aligns more with what was proposed originally in https://github.com/nizamyusli/CRABClient/commit/fdbb0701df3b57818c9b128a66dd723b73f81184

mmascher:
Hi Eric,

yes, logging.DEBUG is 10, logging.INFO=20 and so on. 60 is a value above
logging.CRITICAL (50) that silence everything. I agree to use something
more "userfriendly", a constant noLogInfo should do it.

Cheers,

Marco

PR #4480, opened 2015-05-15
Add function quiet and setLogLevel for crabCommand. Fixes # #4418

nizamyusli:


mmascher:
I say we do not need this. Users can simply do getLoggers(60) and log will only go to files. My 2 cents, we just need to document it and close the github issue.

AndresTanasijczuk:
But in execRaw() there is a "logger, _ = getLoggers(logging.INFO)". Wouldn't that overwrite the getLooger(60) from the user?

AndresTanasijczuk:
Ok, you were right. I need to understand how all this logger thing works, which is not intuitive for me... Closing this pull.

PR #4483, opened 2015-05-19
Separate COMP and CMSSW envs. Fixes #4300

### LINKED ISSUE #4300: Remove CMSSW environment dependencies from crab submit
Currently 'crab submit' needs the CMSSW dependecies because it's useing some environemet variables https://github.com/dmwm/CRABClient/blob/6fa8ddae7d8af9edc7df1718a2773ac1a9c2db7d/src/python/CRABClient/JobType/ScramEnvironment.py#L29-L32 and because it's importing the pset and saving the expanded pickle file https://github.com/dmwm/CRABClient/blob/6fa8ddae7d8af9edc7df1718a2773ac1a9c2db7d/src/python/CRABClient/JobType/CMSSWConfig.py#L61.

All these operations should be in a separate script that we call before doing `scram unset env` and `source /cvmfs/cms.cern.ch/crab3/crab.sh`. The output of this script can be saved somewhere in the task's input directory and then used by the `submit` command.

mmascher:
Provides two new scripts:
  crab.sh: wrapper script that does scram unset env before running crab
commands. Also, if the command is a submit run the bootstrap script
  crab3bootstrap: executes all the operations that needs the CMSSW
environment saving the output to a tmp dir. Returns the tmp directory or
the error message in case of errors.

Backward compatibility is guaranteed since the bootstrapDone() function
is called to check if the clients needs to perform the CMSSW related
operations or the bootstrap script was run.

also refactored many parts of the client. We are providin now:
  flushMemoryLogger function that group the code used to save the output messages that
are stored in memory
  validateSubmitOptions to group the code that in the crab submit
command validate the command line options passed by the user
  setSubmitParserOptions to set the command line options of the client
submit command

### DIFF
--- a/bin/crab
+++ b/bin/crab
@@ -22,9 +22,10 @@ from httplib import HTTPException
 
 from CRABClient.CRABOptParser import CRABOptParser
 from CRABClient import __version__ as client_version
-from CRABClient.ClientUtilities import getAvailCommands, logfilter , uploadlogfile, getLoggers, StopExecution
-from CRABClient.ClientExceptions import ClientException
 from CRABClient.ClientMapping import parametersMapping
+from CRABClient.ClientExceptions import ClientException, FEEDBACKMAIL
+from CRABClient.ClientUtilities import getAvailCommands, logfilter , uploadlogfile, getLoggers, StopExecution, flushMemoryLogger
+
 
 
 class MyNullHandler(logging.Handler):
@@ -74,9 +75,6 @@ class CRABClient(object):
 
             Adapted from Doug Hellmann
             """
-            ## Feedback email address
-            feedbackemail = 'hn-cms-computing-tools@cern.ch'
-
             #adding a file handler
             clientinstance = self
             filehandler = logging.FileHandler(client.logger.logfile)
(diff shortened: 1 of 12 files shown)


mmascher:
Please, discussions on ticket #4300

PerilousApricot:
Hey Marco, I don't think this is the way to have this done. The problem isn't just CMSSW "infecting" CRAB, it's CRAB "infecting" CMSSW as well. If you keep the CRAB env around by default, CMSSW will still have a mixed environment

mmascher:
Hi Andrew,

can you elaborate, where do you see CRAB infecting the CMSSW environment?
The idea is that user will run crab.sh submit blah blah

I am keeping "crab" around because I want to take it slow and distribute
the code as it is now, make sure it is backward compatible. Then at some
point (July release?) make I'll mahe the crab.sh->crab and
crab->somethincelse.py

PerilousApricot:
Because we have users do:

1) set up CMSSW environment
2) set up CRAB environment

So, if a user tries to run CRAB with your new code, things work out fine, we're stripping out the CMSSW stuff (assuming scram unsetenv works right). But, if a user tries to run CMSSW, they're going to be running CMSSW in an environment where CRAB's LD_LIBRARY_PATH is first, which will end up with the same problem of incompatible ABIs as before

mmascher:
Ah ok,

no eventually (hopefully July) the crab.sh script will be included in CMSSW
and removed from here, then people will do:

1) cmsenv
2) crab.sh submit blah (well, we will rename crab.sh to crab)

For June I am keeping it here an I will suggest power users or users who
are haveing problem to directly use the sh script in cvmfs (without
sourcing the CRAB3 environment)

mmascher:
Hi Andrew, I am including my pull in the upcoming client release (due today) because I want to start the tests. After today we start the validation of June, and we are in freeze mode, and I do not want to also miss the June release.

We can include your suggestions in July, at the end of the day I was planning to keep this "hidden" in June and expose it in July anyhow.

PR #4489, opened 2015-05-20
Improve splitting algorithms check in client for PrivateMC and Analysis workflows. Fixes #4246

### LINKED ISSUE #4246: Client should be more descriptive about configuration errors
```
Sending the request to the server
Error contacting the server.
Server answered with: Invalid input parameter
Reason is: Incorrect 'Data.splitting' parameter
Log file is /gpfs22/home/meloam/CRABAPI/ra2taupat_bNAn8/crab_ra2taupat_bNAn8/crab.log
```

I had it lowercase instead of CamelCase, but the client should let the user know what the options are.

AndresTanasijczuk:


AndresTanasijczuk:
I know the error message should come from the server. This should be therefore temporary until we finally move all validation checks to the server with proper customized error messages.

PR #4495, opened 2015-05-27
use proxy for gfal commands if user use --proxy parameter or setting …

juztas:
…env for it


juztas:
More details here: https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/637.html

PR #4498, opened 2015-06-04
add interface directories to tarball, fixes #4497

### LINKED ISSUE #4497: include user header files from scram directory in tarball for ROOT6
see
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/652/1/2/1/1/1/2/1/2/1/1/1/2.html

quoting from there:

For the problem with ROOT6 requiring user code header files, it should
be sufficient to add all the "interface" directories in src, identical
to what is done now for all the "data" directories.

this requires a change in
https://github.com/dmwm/CRABClient/blob/master/src/python/CRABClient/JobType/UserTarball.py

belforte:


belforte:
by the way, seemed simple enough a change that I did not test it. Dunno if this is line with current Crab3 practice.

AndresTanasijczuk:
I did the same change, tested it and works (the interface subdirs in src are added to the tarball recursively).

belforte:
great! I was slowed by needing to learn GIT along the way (thanks @ericvaandering ).

PR #4500, opened 2015-06-09
Improve the way we pass arguments and options to crab commands. Fixes #4180.

### LINKED ISSUE #4180: Improve the way we pass arguments and options to crab commands.
This is a general issue I opened where we can collect different opinions on how we should can improve the way the client parses options and arguments. I think Marco suggested the idea that the client should only parse options and accept no arguments. For example, currently crab submit accepts arguments that are used to override the configuration parameters. These arguments can be passed in one option --overrideconfig="parname1=parvalue1,parname2=parvalue2,...". On the other hand, the option should probably accept only the long format --optname and not the short format -<letter> to avoid the cases in which the users types "-config myconfig.py" (instead of "--config myconfig.py") and the client interprets this as "-c onfig" resulting in an error. I also suggest we do not allow to skip options, i.e. we always require the user to write the option name. For example, do not allow "crab config myconfig.py" but force the user to write "crab config --config myconfig.py".

nizamyusli:


AndresTanasijczuk:
Discussed with Marco, and he prefers to not add the option --overrideconfig since this will disturb users already using the current way of overwriting the config via arguments. What Marco does want to do is to enforce that, in case the --config/-c option was not given, consider that the CRAB config can be in the first argument (but not in the other arguments).

AndresTanasijczuk:
I implemented the above suggestion in #4513

AndresTanasijczuk:
Closing this pull.

Issue #4406, opened 2015-02-18
Improvements for crab status --dryrun


annawoodard:
From @ericvaandering:
1) Do the math for the user and suggest the "correct" splitting to get to our recommended 8 hr jobs.
2) Much more difficult, but allow them to change the splitting in the config file and re-split the task (or I guess delete the current task and replace it transparently)

Also: implement warnings in the case of memory estimate exceeding normal envelope. Also: collect more data about how accurate the estimates are.

mmascher:
For 2, maybe we can edit the splitting args in the oracle DB (and remove the dryrun flag), and the put the task status to `NEW` (instead of `RESUBMITTED`)

annawoodard:
Additional items from @mmascher that can be included in this ticket:

Instead of parsing the dagman input file inside getDryRunOpts, I would like to take the information we need from the task database and/or from some adhoc json file we upload from the TW.

mmascher:
I did a basic testing and it seems to work pretty good! I have one other request for you: here in the TW (*) you are running alternatively the dryrun or the dagman submission, depending on the dryrun flag. do you see anything wrong to change the logic to something like:

```python
handler.addWork( DryRunUploader(config=config, server=server, resturi=resturi, procnum=procnum) )
if task['tm_dry_run'] == 'F':
    handler.addWork( DagmanSubmitter(config=config, server=server, resturi=resturi, procnum=procnum) )
```

meaning we upload the "dryrun sandbox" no matter what, and submit to dagman submitter only if we are not in the "dryrun" mode.

The advantage is that we can add a crab command that runs the "verification code" a-posteriori. Where I am going with it is that if operators could run the dryrun verification code on user's tasks it would be great imho (we can check it the estimation works on users task!)

It should not be a big change, do you think you can work on thet this/next week?

(*)
https://github.com/dmwm/CRABServer/commit/36aa8c4c78500036de7fcad4092eb4a8cc5b8fea#diff-ff551d45cfbad36f9969f573a4176a2dL142

mmascher:
Actually, a possibility is that the `crab submit --dryrun` run only until `Task consists of 5 jobs to process 10 lumis` (*) and the other command does the verification (it saves some time to people only interested in the splitting parameter.. well, maybe people can just hit CTRL-C in that case, but we have to print a message about that.. I don't know :))

(*)
```
[mmascher@lxplus0163 wf]$ crab submit --dryrun
Will use configuration file crabConfig.py.
[...]

PR #4519, opened 2015-07-21
Dry run tweaks

annawoodard:
Collecting the suggestions on #4406:

> Do the math for the user and suggest the "correct" splitting to get to our recommended 8 hr jobs.

Done in 9ad18a2

> Much more difficult, but allow them to change the splitting in the config file and re-split the task (or I guess delete the current task and replace it transparently)

I vote we skip this for now. I don't think the extra complication to the code is worth the slight convenience (in fact, I think it might cause confusion.)

> Implement warnings in the case of memory estimate exceeding normal envelope.

Done in 9ad18a2

> Instead of parsing the dagman input file inside getDryRunOpts, I would like to take the information we need from the task database and/or from some adhoc json file we upload from the TW.

I've modified the function to take the job number as an input parameter instead as an intermediate measure.

> we upload the "dryrun sandbox" no matter what, and submit to dagman submitter only if we are not in the "dryrun" mode.

Done in dmwm/CRABServer@b5a29f3

> Allow user to skip the trial run if they only care about the splitting

I added a ‘skip-estimates’ command-line argument in 9ad18a2.

> Collect more data about how accurate the estimates are.

I added a loop to increase the number of events for psets with low time / event in  9ad18a2 based on their total runtime. Here are typical times for a simulation pset:

Events | AvgEventTime | actual runtime | dryrun estimated runtime (1280 events) | % error (1280 events) 
------------ | ------------ | ------------- | ------------ | -------------
10 | 7.3 | 124.2	 | 9392.9 | -15%
20 | 7.6 | 211.2 | 9743.2 | -19%
40 | 5.4 | 265.9	 | 6855.8 | 16%
80 | 6.5 | 572.7	 | 8365.7 | -3%
160 | 6.5 | 1110.8 | 8312.0 | -2%
320 | 6.3 | 2050.8 | 8001.9 | 2%
640 | 6.5 | 4195.2 | 8276.3 | -1%
1280 | 6.3	 | 8160.6 | 8111.5 | 1%

I've tested that the algorithm executes one 10-event file run for the pset above (~15% error). For a faster-running slimming pset (AvgEventTime ~0.008 s / event), the algorithm executes two iterations, stopping at ~6000 events with a total runtime of ~45 seconds, corresponding to ~25% error. To reduce this error, we could increase the trial time limit. There might also be some more sophisticated ways of doing this-- subtracting the file open time, for example, if someone knows which quantity in the FJR is the correct one-- but I think this is probably close enough for our purposes (and I added an explicit warning about possible issues with the estimates in 9ad18a2.)


src/python/CRABClient/ClientUtilities.py
```diff
@@ -619,8 +619,14 @@ def setSubmitParserOptions(parser):
                            dest = 'dryrun',
                            default = False,
                            action = 'store_true',
-                           help = "Do not actually submit the task; instead, return how many jobs this task would create, along with processing time estimates.")
+                           help = "Do not actually submit the task; instead, return how many jobs this task would create, "\
+                                  "along with processing time and memory consumption estimates.")
 
+    parser.add_option('--skip-estimates',
```
mmascher: @annawoodard , can you document this in the twiki for the commands (I'll let @AndresTanasijczuk tell you which is the right URL :)).

src/python/CRABClient/Commands/submit.py
```diff
@@ -333,57 +332,109 @@ def executeTestRun(self, filecacheurl):
...
+            totalJobSeconds = 0
+            maxSeconds = 25
+            while totalJobSeconds < maxSeconds:
+                opts = getCMSRunAnalysisOpts('Job.submit', 'RunJobs.dag', job=1, events=events)
+
+                s = subprocess.Popen(['sh', 'CMSRunAnalysis.sh'] + opts, env=env, stderr=subprocess.PIPE, stdout=subprocess.PIPE)
+                out, err = s.communicate()
+                self.logger.debug(out)
+                if s.returncode != 0:
+                    raise ClientException('Dry run failed to execute local test run:\n StdOut: %s\n StdErr: %s' % (err, out))
```
mmascher: Invert `out` and `err` please

mmascher:
The pull looks good to me, I added a few minor comments. I will merge this and try to run it once you update it. thanks

AndresTanasijczuk:
Please add documentation about `--skip-estimates` in https://twiki.cern.ch/twiki/bin/view/CMSPublic/CRAB3Commands#crab_submit_dryrun

PR #4520, opened 2015-07-21
Calculate double lumis on the output files instead of on the input files. Fixes #4469

### LINKED ISSUE #4469: Fix how report calculates double lumis: it does not work if the input dataset has split lumis
See https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/553/1.html

If the input dataset has lumi split across files then the crab report will report double lumis. The reason is that report is looking at input files to calculate this. Therefore, if two input files of the same job has the same lumi the report will detect it and report it as a duplicated lumi even if there is only one output file

AndresTanasijczuk:


AndresTanasijczuk:
@mmascher I implemented what we talked last time. Can you review again?

mmascher:
Sounds good to me.

Issue #4528, opened 2015-07-29
check if process.source is in pset


juztas:
See https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/803/1/2/1/1/1/2/1/1/1.html

AndresTanasijczuk:
This is only needed for 'Analysis' workflows, right?

AndresTanasijczuk:
Actually I submitted a PrivateMC task without defining process.source and it fails with that same error. So I guess we have to do the check for both, 'Analysis' and 'PrivateMC'.

AndresTanasijczuk:
But should CRAB do such kind of validation of the CMSSW config at all?

bbockelm:
I think it's reasonable to have CRAB do validation if it's straightforward to implement, has no false positives, and based on support incidents.

ericvaandering:
+1 Of course, no check should be done for scriptExe workflows.

mmascher:
Hi,

just few comments since I was not convinced about this.

I see that the error is there also for montecarlo as well. I took this PSet: https://raw.githubusercontent.com/dmwm/CRABServer/master/test/templates/config/psets/pset_tutorial_MC_generation.py

and commented out this line:

```
#process.source = cms.Source("EmptySource")
``` 

The pset loads correctly when it is imported but `process.source` is `None`:

```
>>> import pset_tutorial_MC_generation
>>> print pset_tutorial_MC_generation.process.source
None
```

And cmsRun fails miserably:

```
----- Begin Fatal Exception 29-Jul-2015 20:53:58 CEST-----------------------
An exception of category 'Configuration' occurred while
   [0] Constructing the EventProcessor
Exception Message:
There must be exactly one source in the configuration.
It is missing (or there are sufficient syntax errors such that it is not recognized as the source)
----- End Fatal Exception -------------------------------------------------
```

So I would say that the fix definitely easy (an `if self.fullConfig.process.source is not None` here https://github.com/dmwm/CRABClient/blob/master/src/python/CRABClient/JobType/CMSSWConfig.py#L60) and I do not think there are false positive if we excluded scriptExe: it seems there is no way cmsRun can succed unless we set an input source.

Sorry for the long comment, but I wanted to document everything also for the benefit of @belforte who Justas said was not convinced by this as well.

belforte:
let me explain my postion, then you do as you like.
[...]

PR #4537, opened 2015-08-04
Make sure process.source exists in CMSSW config.

AndresTanasijczuk:


AndresTanasijczuk:
This is my proposal for https://github.com/dmwm/CRABClient/issues/4528 in case a validation has to be done in CRAB.

PR #4539, opened 2015-08-06
Make JobType.psetName not mandatory for scriptExe.

AndresTanasijczuk:
This is the client part for https://github.com/dmwm/CRABServer/issues/4922


ericvaandering:
test this please

ericvaandering:
test this please

belforte:
@mmascher I can't easily figure out the status of this. IIUC it can't be cleanly merged so would need quite some work to include it. Can you tell if there's useful code to rescue or it was simply an attempt to improve messaging in some corner case. We do not have any obvious need to address in the scriptExe area, it is possible that the original problem was already solved by some other change.
I am very strongly inclined to close.

belforte:
so many years since it was proposed, no burning need for it, can't be cleanly merged and tested... let's close, if needs arise it would easier to do from scratch in current code.

Issue #4522, opened 2015-07-22
Fixes for crab report lumi files


AndresTanasijczuk:
Issue started with this HN: https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/773.html

Discussed in CRAB3 meeting and agreed on the following points:
1) Make crab report work for LumiBased splitting first. Then work on the case of FileBased splitting (in particular input files not in DBS) where crab report should show info about processed/not-processed files instead of lumis.
2) The missing lumis file should be: input lumis AND lumi-mask AND run-range - processed lumis. Another possibility would be to use the lumis corresponding to failed jobs. This is important in case one wants to submit a recovery task before the original task finishes processing (e.g. because crab kill doesn't work and don't want to wait for tails of jobs to finish).

Proposal: crab report should return lumi files for:
a) The input dataset (from DBS, no matter of --dbs option; take the lumis from the data discovery step).
b) The processed lumis in the jobs that finished successfully (including transfer, but not necessarily publication, since publication can anyway be resubmitted -by an operator-). This should be taken from the lumis in input files as written in the FJR, as currently done when --dbs='no'.
c) The non processed lumis = input AND lumi-mask AND run-range - processed = a) AND lumi-mask AND run-range - b)
d) The lumis in the published output datasets (one file for each published dataset).
e) The duplicate lumis in the output. We don't need to use DBS; it is enough to look for duplicate lumis in b).
f) The duplicate lumis in the input, i.e. duplicate lumis in a) ? Even if CRAB can deal with them, it would be good to give this to users IMHO in case they were not expecting it. Especially for USER datasets.

ericvaandering:
To point 1) if the input files contain lumi information, we should report it. People will still want to take that file and calculate the luminosity of their analysis. Of course, you can't tell what is "missing" in this case.

AndresTanasijczuk:
@belforte I thought we have agreed on this issue, but seems we didn't. For example, we don't want to do a) and f) right?
About
c) The non processed lumis = input AND lumi-mask AND run-range - processed
It should actually be 
c) The non processed lumis = input AND lumi-mask AND run-range AND totalUnits cut - processed,
right? I.e. what CRAB was asked to process - what it finished processing. And we should not construct (input AND lumi-mask AND run-range AND totalUnits cut) ourselves, but take it from the files in run_and_lumis.tar.gz file, right?

belforte:
[...]

PR #4558, opened 2015-09-21
Add user function getLumiListInValidFiles

AndresTanasijczuk:
This PR is for https://github.com/dmwm/CRABServer/issues/4982 Ohhh, I didn't see Stefano's comment that he volunteered to do this.. sorry...

### DIFF
--- a/src/python/CRABClient/UserUtilities.py
+++ b/src/python/CRABClient/UserUtilities.py
@@ -10,10 +10,15 @@ import traceback
 from urlparse import urlparse
 from ast import literal_eval
 
+## DBS dependencies
+from dbs.apis.dbsClient import DbsApi
+
 ## WMCore dependencies
 from WMCore.Configuration import Configuration
+from WMCore.DataStructs.LumiList import LumiList
 
 ## CRAB dependencies
+from CRABClient.ClientUtilities import DBSURLS
 from CRABClient.ClientExceptions import ClientException, UsernameException, ProxyException
 
 
@@ -150,6 +155,42 @@ def getFileFromURL(url, filename = None, proxyfilename = None):
     with open(filename, 'w') as f:
         f.write(filestr)
     return filename
+
+
+def getLumiListInValidFiles(dataset, dbsurl = 'phys03'):
+    """
+    Get the runs/lumis in the valid files of a given dataset.
+
+    dataset: the dataset name as published in DBS
+    dbsurl: the DBS URL or DBS prod instance
+
+    Returns a LumiList object.
+    """
+    dbsurl = DBSURLS['reader'].get(dbsurl, dbsurl)
+    dbs3api = DbsApi(url=dbsurl)
+    try:
+        files = dbs3api.listFileArray(dataset=dataset, validFileOnly=0, detail=True)
+    except Exception as ex:
+        msg  = "Got DBS client error requesting details of dataset '%s' on DBS URL '%s': %s" % (dataset, dbsurl, ex)
+        msg += "\n%s" % (traceback.format_exc())
+        raise ClientException(msg)
+    if not files:
+        msg = "Dataset '%s' not found in DBS URL '%s'." % (dataset, dbsurl)
+        raise ClientException(msg)
+    validFiles = [f['logical_file_name'] for f in files if f['is_file_valid']]
+    blocks = set([f['block_name'] for f in files])
+    runLumiPairs = []
+    for blockName in blocks:
+        fileLumis = dbs3api.listFileLumis(block_name=blockName)
+        for f in fileLumis:
+            if f['logical_file_name'] in validFiles:
+                run = f['run_num']
+                lumis = f['lumi_section_num']
+                for lumi in lumis:
+                    runLumiPairs.append((run,lumi))
+    lumiList = LumiList(lumis=runLumiPairs)
+    
+    return lumiList
 
 
 def getLoggers():
--- a/bin/crab3bootstrap
+++ b/bin/crab3bootstrap
@@ -16,7 +16,6 @@ import sys
 import json
 import tempfile
 import traceback
-from optparse import OptionParser
 
 from ServerUtilities import FEEDBACKMAIL
 
@@ -78,7 +77,7 @@ def getTmpDir():
     return tempName
 
 
-def getConfigName(logger):
+def getConfigName():
     """ Get the configuration filename from the command line options
     """
     ## The client has a "two level" parsing. One global (for example gets the --debug option and the command name),
@@ -90,7 +89,7 @@ def getConfigName(logger):
     opt, args =  parser.parse_args()
     args = args or []
     opt, cmdargs = optParser.parse_args(args[1:])
-    validateSubmitOptions(opt, args, logger)
+    validateSubmitOptions(opt, args)
 
     return opt.config
 
@@ -117,7 +116,7 @@ if __name__ == '__main__':
         setConsoleLogLevelVar(LOGLEVEL_MUTE)
         tblogger, logger, memhandler = None, None, None
         tblogger, logger, memhandler = initLoggers()
-        cfgName = getConfigName(logger)
+        cfgName = getConfigName()
         config = getConfig(cfgName)
         cmsswCfg = CMSSWConfig(config=config, userConfig=config.JobType.psetName, logger=logger)
         valid, msg = cmsswCfg.validateConfig()
--- a/src/python/CRABClient/ClientUtilities.py
+++ b/src/python/CRABClient/ClientUtilities.py
@@ -8,7 +8,6 @@ import copy
 import datetime
 import logging
 import logging.handlers
-import string
 import time
 import pkgutil
 import sys
@@ -280,7 +279,7 @@ def getPlugins(namespace, plugins, skip):
     return modules
 
 
-def addPlugin(pluginpathname, pluginname = None):
+def addPlugin(pluginpathname):
     """
     _addPlugin_
 
@@ -635,7 +634,7 @@ def setSubmitParserOptions(parser):
                            action = 'store_true',
                            help = "When executing a dry run, skip the processing time and memory consumption estimates.")
 
-def validateSubmitOptions(options, args, logger):
+def validateSubmitOptions(options, args):
     """ If no configuration file was passed as an option, try to extract it from the first argument.
         Assume that the arguments can only be:
             1) the configuration file name (in the first argument), and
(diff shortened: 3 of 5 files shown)


src/python/CRABClient/UserUtilities.py
```diff
@@ -152,6 +159,43 @@ def getFileFromURL(url, filename = None, proxyfilename = None):
...
+    """
+    Get the runs/lumis in the valid files of a given dataset.
+
+    dataset: the dataset name as published in DBS
+    dbsurl: the DBS URL or DBS prod instance
+
+    Returns a LumiList object.
+    """
+    from dbs.apis.dbsClient import DbsApi
+    from WMCore.DataStructs.LumiList import LumiList
```
ericvaandering: Please put imports at the top of the file unless there is some reason not to.

belforte:
thanks @AndresTanasijczuk no reason to be sorry !

AndresTanasijczuk:
Moved the imports to module level as suggested by @ericvaandering

AndresTanasijczuk:
Documented in https://twiki.cern.ch/twiki/bin/view/CMSPublic/CRAB3UserFunctions#get_LumisInDatasetFromDBS

belforte:
thanks. I do not understand the example, maybe I have a wrong expectation about the output of crab report. Let's come to that once release is out.

AndresTanasijczuk:
I added the possibility to save the result to a file (in JSON). 

If we have this function, I guess there is no need for documenting how to use the DBS client to get the lumis of a dataset, right?

belforte:
right.

AndresTanasijczuk:
@belforte About the example in the documentation, what I had in mind is the case in which one uses a lumi-mask that spans a larger run-range than the dataset itself. For example, one can have a lumi-mask for the whole LHC RunN, but the dataset is split in RunNA, RunNB, etc. In such a case, if one uses `crab report` to get the analyzed/missing lumis, and assuming crab report doesn't query DBS (`crab report --dbs='no'`), the `missingLumiSummary.json` file will be constructed as the lumi-mask minus the analyzed lumis. I guess you see the point now...
Anyway, I removed the example, because it's probably not much instructive (but if you want I can put it back). I left instead the pointer to the relevant example that motivated the introduction of this function: dealing with a growing dataset.

belforte:
@AndresTanasijczuk let's discuss the example(s) in the original issue, which has a better subject
https://github.com/dmwm/CRABServer/issues/4982

But I think this is the wrong thing to do: 

`the missingLumiSummary.json file will be constructed as the lumi-mask minus the analyzed lumis. `

I hope you are misunderstanding the code. If not, we have to change the code. I have not looked yet. If there's any chance that indeed current code fills the report in the wrong way, we should open an ad hoc issue.

belforte:
@AndresTanasijczuk what about renaming:
getLumisInDatasetFromDBS --> getValidLumisInDataset
since Lumi information is in DBS only by design and make it clear the main difference with DBS API (at least to expert) as not only a format change, but also here we take one flexibility out.
I have played a bit also with the name
getLumilistFromDataset(dateset=None, url=global, validOnly=True)

what do you think ? I do not see the usefulness to pass an output file to this, e.g. alredy the example you have now
https://twiki.cern.ch/twiki/bin/view/CMSPublic/CRAB3FAQ#Dealing_with_a_growing_input_dat
indicates to use directly filteredLumiMask.writeJSON to save a new lumiMask, and indeed there is no point in there to save new_lumi_mask.json which IMHO is a confusing name and concept. THe list of lumis in the input dataset is NOT a lumimask.

By the way that example may have a typo... it seems to use 'new_lumi_mask.json' without having ever create it

AndresTanasijczuk:
@belforte About 
`the missingLumiSummary.json file will be constructed as the lumi-mask minus the analyzed lumis.`
that's what the code does now if one chooses `--dbs='no'`. There is an issue to fix the files from `crab report`: https://github.com/dmwm/CRABClient/issues/4522

belforte:
since we agreed that crab report  needs fixing in #4522 I would not give to users examples based on the bad code. But let's keep that separate.
In here let's stick to how to name the new function. I like to to use a name which says what it does i.e. return (valid only?) lumis in Lumilist format.

AndresTanasijczuk:
@belforte About renaming the function, feel free to propose the best name you can find.
We have to be careful with the validOnly, because for DBS API if one chooses validOnly=True that automatically means dataset must be VALID (and also PRODUCTION I think is allowed).
About of saving result in output file, I am ok with removing that.
About the example, what I called `new_lumi_mask.json` is the growed lumi-mask from Physics Validation. I create a LumiList object from it and call it `newLumiMask`. I get the lumis in the dataset publisehd by TaskA with the new function and call it `taskALumis`. Then a create `filteredLumiMask = newLumiMask - taskALumis` which I think is ok to call it a "lumi-mask" (but name can be improved of course), and then I save it in a file `my_lumi_mask.json` (again name can be improved).

belforte:
anyhow, the text in
https://twiki.cern.ch/twiki/bin/view/CMSPublic/CRAB3FAQ#Dealing_with_a_growing_input_dat
appears a bit verbose, but correct to me.
The example may need a bit work to (appear clearly to) be consistent with it, that's all.

belforte:
e.g. `new_lumi_mask.json` should be defined in the twiki, non in your comment above :-)
but improving documentation is never-ending work

AndresTanasijczuk:
Sure, feel free to edit the twiki.

AndresTanasijczuk:
Can you put here (whenever you are done) the final version you would like to have for the function (the whole code) ?

belforte:
I edited the twiki. Actuallty added what I think will do leaving current stuff below. Please review. If you agree we can also remove all text below the IMPORTANT NOTE

The function name in the twki can be changed after we settle on the one here. But not today. Have a couple of meetings

ericvaandering:
The Twiki in its current form looks OK to me, but I'm not sure all the stuff after IMPORTANT NOTE adds anything either. Best to keep instructions short, accurate, and informative.

belforte:
@AndresTanasijczuk hoping I understood correctly your comment https://github.com/dmwm/CRABClient/pull/4558#issuecomment-142844825 , here it is
```python
def lumiListInDataset(dataset, dbsurl = 'phys03', validFileOnly = True):
    """
    Get the runs/lumis in the valid files of a given dataset.
    dataset: the dataset name as published in DBS
    dbsurl: the DBS URL or DBS prod instance
    Returns a LumiList object.
    """
    dbsurl = DBSURLS['reader'].get(dbsurl, dbsurl)
    dbs3api = DbsApi(url=dbsurl)
    try:
        files = dbs3api.listFileArray(dataset=dataset, validFileOnly=validFileOnly, detail=True)
    except Exception as ex:
        msg  = "Got DBS client error requesting details of dataset '%s' on DBS URL '%s': %s" % (dataset, dbsurl, ex)
        msg += "\n%s" % (traceback.format_exc())
        raise ClientException(msg)
    if not files:
        msg = "Dataset '%s' not found in DBS URL '%s'." % (dataset, dbsurl)
        raise ClientException(msg)
    validFiles = [f['logical_file_name'] for f in files if f['is_file_valid']]
    blocks = set([f['block_name'] for f in files])
    runLumiPairs = []
    for blockName in blocks:
        fileLumis = dbs3api.listFileLumis(block_name=blockName)
        for f in fileLumis:
            if f['logical_file_name'] in validFiles:
                run = f['run_num']
                lumis = f['lumi_section_num']
                for lumi in lumis:
                    runLumiPairs.append((run,lumi))
    lumiList = LumiList(lumis=runLumiPairs)

    return lumiList
```

ericvaandering:
I think you can simplify (and speed up) this just by using runsAndLumis form of the LumiList constructor.

See https://github.com/ericvaandering/WMCore/blob/9dc5322c460556dfa3e5b13bf5f3afdb836f5c70/src/python/WMCore/DataStructs/LumiList.py#L452

belforte:
unfortunately DBS does not give lumis in that format, but returns: List of dictionaries containing the following keys (lumi_section_num, logical_file_name, run) One dictionary for each lumi, many dictionaries for every run.

ericvaandering:
This right here: 

                for lumi in lumis:
                    runLumiPairs.append((run,lumi))


you can eliminate. You may have to do lumiList += LumiList(runsAndLumis = {run: lumis}) though.

belforte:
well, this is Andres code. Not to wash my hands... but maybe he should be the one to comment ?

indeed there's no use in looping on lumis, but looking at code to be executed to convert runsAndLumis format into compact and OR the compact at each  += , I am not sure it is worth to do it for each lumi. Converting each pair (r,l) into an extensione of {r:[l,...,l]} may be as much work as doing it for the full list at the end.

AndresTanasijczuk:
We can change to what Eric proposes. It was just a matter of (my) taste to define the LumiList object only once at the end with the final list of runs/lumis.

ericvaandering:
Lots of ways to do this and you should do whatever works best for you (except re-inventing the wheel). Maybe look at some performance tests for extreme examples.

AndresTanasijczuk:
@belforte You add validFileOnly as input argument. Can you explain what is the purpose? What if the user set validaFileOnly=False? The code is anyway later filtering out invalid files, which I think is want we want to do, given that CRAB only analyzes valid files.

belforte:
@AndresTanasijczuk you are right. I saw validFileOnly used in listFileArray and thought that instead of hardcoding, one could get it from the caller. But as you say this is not consistent now. So better to do:
def ValidLumiListFromDataset(dataset, dbsurl = 'phys03')
or maybe this is more clear
def getLumiListForValidFiles(dataset, dbsurl='phys03')

OTOH I do not understand why we need to get list of vaild files via listFIleArray if later you check every file_is_valid

AndresTanasijczuk:
`validFileOnly` in DBS client does not mean what one would naively think. If one set validFileOnly=1 in DBS client, this will automatically discard datasets with status other than VALID (and I don't remember if also PRODUCTION is allowed) and only then filter out invalid files. While in CRAB we only care about the status of files, not dataset. That's why in the server we also use validFileOnly=0 and then we filter out invalid files:
* https://github.com/dmwm/CRABServer/blob/master/src/python/TaskWorker/Actions/DBSDataDiscovery.py#L118
* https://github.com/dmwm/CRABServer/blob/master/src/python/TaskWorker/Actions/DataDiscovery.py#L43-L46

belforte:
right. I was vaguely remember and was trying to look it up.
thanks. Sorry.
pff... we need to bug Yuyi to at least document this :-(

AndresTanasijczuk:
Ok, so the only change at the end is the function name and revisit what Eric proposed.

AndresTanasijczuk:
Using this test function:
```
def getLumisInDatasetFromDBS(dataset, dbsurl = 'phys03'):
    """
    Get the runs/lumis in the valid files of a given dataset.

    dataset: the dataset name as published in DBS
    dbsurl: the DBS URL or DBS prod instance

    Returns a LumiList object.
    """
    dbsurl = DBSURLS['reader'].get(dbsurl, dbsurl)
    dbs3api = DbsApi(url=dbsurl)
    try:
        files = dbs3api.listFileArray(dataset=dataset, validFileOnly=0, detail=True)
    except Exception as ex:
        msg  = "Got DBS client error requesting details of dataset '%s' on DBS URL '%s': %s" % (dataset, dbsurl, ex)
        msg += "\n%s" % (traceback.format_exc())
        raise ClientException(msg)
    if not files:
        msg = "Dataset '%s' not found in DBS URL '%s'." % (dataset, dbsurl)
        raise ClientException(msg)
    validFiles = [f['logical_file_name'] for f in files if f['is_file_valid']]
    blocks = set([f['block_name'] for f in files])
    blocksLumis = {}
    for blockName in blocks:
        fileLumis = dbs3api.listFileLumis(block_name=blockName)
        blocksLumis[blockName] = fileLumis

    # method 1
    runLumiPairs = []
    method1_start = datetime.datetime.now()
    print "starting method 1 at %s" % method1_start
    for blockName, fileLumis in blocksLumis.iteritems():
        for f in fileLumis:
            if f['logical_file_name'] in validFiles:
                run = f['run_num']
                lumis = f['lumi_section_num']
                for lumi in lumis:
                    runLumiPairs.append((run,lumi))
    lumiList1 = LumiList(lumis=runLumiPairs)
    method1_end = datetime.datetime.now()
    print "finished method 1 at %s" % method1_end
    # method 2
    lumiList2 = LumiList()
    method2_start = datetime.datetime.now()
    print "starting method 2 at %s" % method2_start
    for blockName, fileLumis in blocksLumis.iteritems():
        for f in fileLumis:
            if f['logical_file_name'] in validFiles:
                run = f['run_num']
                lumis = f['lumi_section_num']
                lumiList2 += LumiList(runsAndLumis={run: lumis})
    method2_end = datetime.datetime.now()
    print "finished method 2 at %s" % method2_end

return
```

I ran it over dataset `/MinimumBias/Run2015A-v1/RAW` (https://cmsweb.cern.ch/das/request?input=dataset%3D%2FMinimumBias%2FRun2015A-v1%2FRAW&instance=prod%2Fglobal) and produced this output:
```
>>> from CRABClient.UserUtilities import getLumisInDatasetFromDBS
>>> getLumisInDatasetFromDBS('/MinimumBias/Run2015A-v1/RAW', dbsurl='global')
starting method 1 at 2015-09-25 18:54:37.590492
finished method 1 at 2015-09-25 18:54:37.717539
starting method 2 at 2015-09-25 18:54:37.717621
finished method 2 at 2015-09-25 18:54:43.748537
>>>  
```
Just in case I inverted the order of method 1 and 2 in the test function and got this output.
```
>>> from CRABClient.UserUtilities import getLumisInDatasetFromDBS
>>> lumiList1, lumiList2, start1, end1, start2, end2 = getLumisInDatasetFromDBS('/MinimumBias/Run2015A-v1/RAW', dbsurl='global')
starting method 2 at 2015-09-25 18:58:02.906293
finished method 2 at 2015-09-25 18:58:09.137926
starting method 1 at 2015-09-25 18:58:09.137999
finished method 1 at 2015-09-25 18:58:09.261181
```
So method 1 of looping over lumis, building a list with all (run, lumi) pairs and calling LumiList once at the end was overwhelmingly much faster than calling LumiList for each {run: lumis}.
And of course I checked that the two methods gave the same result.
Is that convincing? Should we keep the function as it is then?

AndresTanasijczuk:
@belforte Assuming the above test was enough, it only remains to decide the function name, which I will let at your hands.

ericvaandering:
Yes. Go with what performs.

belforte:
test was enough. Thanks for running it.
as name goes... let's settle on

getLumiListInValidFiles(dataset,url)

the concepts of non-valid datasets and files are currenly
confused and confusing enough that I am for being clear
to the point of being pedantic

AndresTanasijczuk:
Function name updated to getLumiListInValidFiles.

AndresTanasijczuk:
@mmascher As for the function, I think this PR is ready. I am not sure if you agree to move DBSURLS to ClientUtilities.py. For sure it has to be moved away from submit.py, but not sure where is the best place to put it. Maybe ServerUtilities.py? Or ClientMapping.py?

belforte:
Since client does not need it for talking to the server, ClientUtilities seems a fine place to me

ericvaandering:
test this please

AndresTanasijczuk:
test this please

AndresTanasijczuk:
The last pylint check says there is an error in crab3boostrap (E1120) but I don't see such error. And googling I found some people complaining about false positives of E1120.

ericvaandering:
It's possible it somehow has the old signature for validateSubmitOptions which did require logger

mmascher:
Merging, I have read the whole conversaion only now. I also had some evidence that accumulating lumis in a set and creating the LumiList object only once was faster (see my change here https://github.com/dmwm/CRABClient/commit/975e53813ce50d5297e21d8959b0a56e4254bac7)

PR #4561, opened 2015-10-05
Only consider failed short log transfers when error is not expected.

AndresTanasijczuk:


belforte:
can you add one line of comment with:
# status 404 means file not found, see http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

belforte:
wow.. what an effect a simple '#' had !

AndresTanasijczuk:
@belforte Added comments.

PR #4567, opened 2015-10-13
Change hasattr -> getattr for attributes that are now initialized to None.

AndresTanasijczuk:


belforte:
was a change to client pushed to CVMFS w/o testing ?

PR #4569, opened 2015-10-14
Remove check of no input dataset in PrivateMC.

AndresTanasijczuk:
Moving this check to the server.


mmascher:
The check that the inputdataset is empty for privateMC is done in the server now: https://github.com/dmwm/CRABServer/pull/5042

PR #4573, opened 2015-11-09
Fixes for crab report command. Fixes #4522

### LINKED ISSUE #4522: Fixes for crab report lumi files

AndresTanasijczuk:
On top of what is discussed in #4522, I also: 
* Removed --usedbs option, because in the report the output/input datasets lumis are always returned (the input datasets lumis from a file saved at data discovery time).
* Added --failed option (feel free to suggest a better name): accepts a comma-separated list of jobs states, e.g. `--failed=failed,killed`, in which case the report will return as the not-processed lumis the lumis that all jobs in those states should have processed. Otherwise the default for not-processed lumis is = lumis in run_and_lumis.tar.gz - processed lumis.

This PR has a counterpart in the server: https://github.com/dmwm/CRABServer/pull/5046. The server PR is backward compatible with current client and with old tasks.

### DIFF
--- a/src/python/CRABClient/JobType/BasicJobType.py
+++ b/src/python/CRABClient/JobType/BasicJobType.py
@@ -5,8 +5,10 @@ Conventions:
  2) a plug-in needs to implement mainly the run method
 """
 from ast import literal_eval
+
 from WMCore.Configuration import Configuration
 from WMCore.DataStructs.LumiList import LumiList
+
 from CRABClient.ClientExceptions import ConfigurationException
 
 
@@ -51,9 +53,9 @@ class BasicJobType(object):
 
 
     @staticmethod
-    def mergeLumis(inputdata, lumimask):
+    def mergeLumis(inputdata):
         """
-        Computes the processed lumis, merges if needed and returns the compacted list (called when usedbs=no).
+        Computes the processed lumis, merges if needed and returns the compacted list.
         """
         mergedLumis = set()
         #merge the lumis from single files
@@ -63,24 +65,32 @@ class BasicJobType(object):
                     for lumi in lumis:
                         mergedLumis.add((run,int(lumi))) #lumi is str, but need int
         mergedLumis = LumiList(lumis=mergedLumis)
-        diff = LumiList(compactList=lumimask) - mergedLumis
-        return mergedLumis.getCompactList(), diff.getCompactList()
+        return mergedLumis.getCompactList()
 
 
     @staticmethod
-    def subtractLumis(input, output):
-        """
-        Computes the processed lumis, merges from the DBS reuslts (called when usedbs=yes).
-        """
-        out = LumiList(runsAndLumis=output)
-        in_ = LumiList(runsAndLumis=input)
-        diff = in_ - out
-        return out.getCompactList(), diff.getCompactList()
+    def intersectLumis(lumisA, lumisB):
+        result = LumiList(compactList=lumisA) & LumiList(compactList=lumisB)
+        return result.getCompactList()
 
 
     @staticmethod
-    def getDoubleLumis(lumisDict):
-        #calculate lumis counted twice
+    def subtractLumis(lumisA, lumisB):
+        result = LumiList(compactList=lumisA) - LumiList(compactList=lumisB)
+        return result.getCompactList()
+
+
+    @staticmethod
+    def getDuplicateLumis(lumisDict):
+        """
+        Get the run-lumis appearing more than once in the input
+        dictionary of runs and lumis, which is assumed to have
+        the following format:
+            {
+            '1': [1,2,3,4,6,7,8,9,10],
+            '2': [1,4,5,20]
+            }
+        """
         doubleLumis = set()
         for run, lumis in lumisDict.iteritems():
             seen = set()
--- a/src/python/CRABClient/JobType/BasicJobType.py
+++ b/src/python/CRABClient/JobType/BasicJobType.py
@@ -6,7 +6,6 @@ Conventions:
 """
 from ast import literal_eval
 
-from WMCore.Configuration import Configuration
 from WMCore.DataStructs.LumiList import LumiList
 
 from CRABClient.ClientExceptions import ConfigurationException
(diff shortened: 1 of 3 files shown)


belforte:
indeed --failed=failed is not intuitive.
I think I understand the spirit (tell me what I have to recover even if task
is not over yet and I have pending/running jobs).
So, effectively you want the list of lumis from jobs in a given set of states, correct ?
What about:
  crab report --statuses=failed,killed
and possibly only produce one file in that case, the lumis in those jobs.

AndresTanasijczuk:
Yes, your interpretation of the --failed option is correct. And yes, I think --statuses is better. 
About producing only one lumi file when --statuses=.... is given, if you are sure that this is what we should do, I have no problem to implement it. Should I also suppress the messages about number of events/files processed/written in that case?

AndresTanasijczuk:
I added back the --dbs option. When this option is specified, the not-processed lumis are calculated as: lumis in input dataset in DBS - lumis on all output datasets in DBS. Notice that a task can publish in more than one output dataset. With the above definition of not-processed lumis, if a lumi from the input dataset was not published in ALL the output datasets, it will end up being part of the not-processed lumis.

belforte:
given all the additional complexities:
only consider failed/killed jobs..
multiple datasets with possibly not same lumi list...
I fear we are trying to get too much for a single command, and
the meaning of "report" is lost.

that said, I do not understand what you want to do here. i.e. what's different
between:
lumis in input dataset in DBS - lumis on all output datasets in DBS.
and
if a lumi from the input dataset was not published in ALL the output datasets, it will end up being 
part of the not-processed lumis

In general, if a job wants to publish in multiple datasets, but CRAB only does
some of them, recovery is going to be tricky.

Maybe all this part about output datasets should be left to
users doing by themselves using lumimask algebra like in
https://twiki.cern.ch/twiki/bin/view/CMSPublic/CRAB3FAQ#Dealing_with_a_growing_input_dat


On 11/24/2015 03:56 PM, Andres Tanasijczuk wrote:
> I added back the --dbs option. When this option is specified, the not-processed lumis are calculated
> as: lumis in input dataset in DBS - lumis on all output datasets in DBS. Notice that a task can
> publish in more than one output dataset. With the above definition of not-processed lumis, if a lumi
> from the input dataset was not published in ALL the output datasets, it will end up being part of
> the not-processed lumis.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dmwm/CRABClient/pull/4573#issuecomment-159292341>.
>

-- 
Stefano Belforte - I.N.F.N.     tel   : +39 040 375-6261 (fax: 375-6258)
Area di Ricerca - Padriciano 99 tel mobile: +39 328 010 7327
34012 TRIESTE TS  - Italy       AIM: stefanobelforte

AndresTanasijczuk:
Changed the --failed option to --statuses and allow only failed/killed for this option.

ericvaandering:
test this please

src/python/CRABClient/JobType/BasicJobType.py
```diff
@@ -63,24 +64,32 @@ def mergeLumis(inputdata, lumimask):
...
-    def subtractLumis(input, output):
-        """
-        Computes the processed lumis, merges from the DBS reuslts (called when usedbs=yes).
-        """
-        out = LumiList(runsAndLumis=output)
-        in_ = LumiList(runsAndLumis=input)
-        diff = in_ - out
-        return out.getCompactList(), diff.getCompactList()
+    def intersectLumis(lumisA, lumisB):
+        result = LumiList(compactList=lumisA) and LumiList(compactList=lumisB)
```
mmascher: I thought this had to be `&`

AndresTanasijczuk:
@belforte Do you remember what we talked about the --recovery option? Recap: We said that it should accept three values: 1) 'notFinished', 2) 'notPublished', 3) 'failed', and that in each case the list of lumis for the recovery task (what we were always calling "missingLumis") should be calculated as:
1) the lumis that the task should have processed (the runsAndLumis tarball) minus what has successfully been processed (up to transfers and filemetadata upload, but not necessarily publication);
2) the lumis in the input dataset minus the published lumis;
3) the lumis that should have been analyzed by failed jobs.
I was thinking if 2) is useful at all, or instead should it be more appropriate to do: the lumis that the task should have processed (the runsAndLumis tarball) minus what has been published. Actually, in that case 1) and 2) will become the same once we integrate the publication as a step of the task that the DAG FINAL node will monitor.

belforte:
@AndresTanasijczuk i would have thought that we agreed already to always do
"what was input" (runlumi.tgz) - "something".
Indeed in 1) and 2) you take the same starting point, but in one case subtract what was successfully staged out, in the other what was published. I.e. we should do what you wrote as

"more appropriate to do: the lumis that the task should have processed (the runsAndLumis tarball) minus what has been published"

Even if we make publication always succeed there will be a usefulness for the two possibilities for the situation where the user does not want to publish the output.

AndresTanasijczuk:
Ok @belforte Done.

PR #4585, opened 2015-12-11
Support options without value in kwargs of crabCommand. Fixes #4577

### LINKED ISSUE #4577: Simplify the way of passing options to a crab command in crabCommand
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/1159.html

We should make passing options to a crab command via crabCommand the most intuitive and simple as possible. Current situation is https://github.com/dmwm/CRABClient/blob/3.3.1511/src/python/CRABAPI/RawCommand.py#L12 which accepts the following possibilities:

* Suppose all options I pass require a value:
1) crabCommand('\<command\>', optname1=optvalue1, optname2=optname2) which uses "kwargs"
The client API then puts all arguments in a list: ['--optname1', optvalue1, '--optname2', optvalue2]
2) crabCommand('\<command\>', '--optname1=optvalue2', '--optname2=optvalue2') which uses "args"
The client API then puts all arguments in a list: ['--optname1=optvalue1', '--optname2=optvalue2']
3) crabCommand('\<command\>', '--optname1', optvalue1, '--optname2', optvalue2) which uses "args"
The client API then puts all arguments in a list: ['--optname1', optvalue1, '--optname2', optvalue2]

* Suppose some options do not require a value, then the user is forced to pass those before the keyword arguments:
1) crabCommand('\<command\>', '--optname1', optname2=optname2) which uses "args" and "kwargs"
The client then puts all arguments in a list: ['--optname2', optvalue2, '--optname1']
2) crabCommand('\<command\>', '--optname1', '--optname2=optvalue2') which uses "args"
The client then puts all arguments in a list: ['--optname1', '--optname2=optvalue2']
3) crabCommand('\<command\>', '--optname1', '--optname2', optvalue2) which uses "args"
The client then puts all arguments in a list: ['--optname1', '--optname2', optvalue2]

In the hypernews issue I pasted above I told the user that I will document examples of how one can pass options to the command in crabCommand. But I am now inclined to not document anything and fix this issue already in December release as I think it is very easy.

I propose to simplify crabCommand to accept a unique format for the options/arguments to the command. For example:

crabCommand(command, argsstr)

So the user will have to call crabCommand with the following syntax:

crabCommand('\<command\>', '--optname1=optvalue1 --optname2 --optname3 optvalue3 arg1 arg2')
We will take cmdargsstr and make a list by just splitting it (using empty space as separator): ['--optname1=optvalue1', '--optname2', '--optname3', 'optvalue3', 'arg1', 'arg2']

So the code I propose is:
```
def crabCommand(command, argsstr=None):
    """ crabComand - executes a given command with certain arguments and returns
                     the raw result back from the client.
    """
    args = argsstr.split() if argsstr else []

    return execRaw(command, args)
```

AndresTanasijczuk:
Documentation of the crabCommand API already updated in https://twiki.cern.ch/twiki/bin/view/CMSPublic/CRABClientLibraryAPI

### DIFF
--- a/src/python/CRABAPI/RawCommand.py
+++ b/src/python/CRABAPI/RawCommand.py
@@ -16,8 +16,12 @@ def crabCommand(command, *args, **kwargs):
     #Converting all arguments to a list. Adding '--' and '='
     arguments = []
     for key, val in kwargs.iteritems():
-        arguments.append('--'+str(key))
-        arguments.append(val)
+        if isinstance(val, bool):
+            if val:
+                arguments.append('--'+str(key))
+        else:
+            arguments.append('--'+str(key))
+            arguments.append(val)
     arguments.extend(list(args))
 
     return execRaw(command, arguments)
--- a/src/python/CRABAPI/RawCommand.py
+++ b/src/python/CRABAPI/RawCommand.py
@@ -3,7 +3,6 @@
         but doesn't want to subprocess.Popen()
 """
 import CRABAPI
-import logging
 
 from CRABClient.ClientUtilities import initLoggers, flushMemoryLogger, removeLoggerHandlers
 


ericvaandering:
test this please

PR #4590, opened 2016-01-26
Fix getoutput for new pycurl_manager

mmascher:


mmascher:
Fixes a bug introduced in a new version WMCore version of the pycurl_manager. This https://github.com/dmwm/WMCore/commit/36a9906ab45b928ddeeea59258cf34672fd2599d should have fixed the bug but there are cases where the parameters are passes as a 2tuple-list instead of a dict.

Issue #4390, opened 2015-02-04
Upload log even if .requestcache file does not exist


juztas:
See https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/387/1/1/2/1/1.html

mmascher:
Actually that should already be an option --file or something like that. Maybe if the .requestcache file is not there (=>failure before the taskdir is set) we can take crab.log?

belforte:
this is labelled as Done. Can it be closed ?

AndresTanasijczuk:
The pull request https://github.com/dmwm/CRABClient/pull/4397 is not merged yet. I use the label "Done" to flag issues that have already a pull request pending.

belforte:
ok. moved to next release then.

emaszs:
Turns out there is an option for uploadlog called --logpath [1] which will allow the log upload without having a .requestcache file, so we probably don't need this. @belforte, can we close this?\

[1] https://github.com/dmwm/CRABClient/blob/master/src/python/CRABClient/Commands/uploadlog.py#L22

belforte:
could we do that step automatically in case .requestcache does not exit ?

mmascher:
Keep in mind that in case the .requestcache is missing and you do not have a taskid, then you only have a `crab.log` file, and to distinguish between different uploads the client adds a timestamp. How does the operator knows this timestamp? The user has to tell him via email, and since this additional iteration is already needed let's just do the `upload --logpath`. So IMHO this is basically useless :)

Basically I do not like https://github.com/nizamyusli/CRABClient/commit/8421a281b8dd896bf2ceb71efe8fc34ad2c5279f#diff-9e49c971373c9c5973d04709a75d96b3R35 . On the other hand (besides this) the patch looks good. Needs a merge or a rebase.

belforte:
should we then print a message telling the user to locate the log himself and use --logpath ?

mmascher:
+1

@emaszs have you trid to see what happens in the case mentioned in the hypernews at the top of the isse? Both before and after applying the patch.

PR #4604, opened 2016-03-16
Fix conflicts for PR #4390, don't upload crab.log without .requestcache.

emaszs:


mmascher:
Looks good to me.

emaszs:
I've added a message for the user as per @belforte 's comment in the original issue [1] as well as an else / return statement because without a .requestcache there's an undefined value, and we don't want to upload the log file anyway.

Let me know if there's a more appropriate way to break out of this function (perhaps return {'result' : {'status' : 'FAILED'}} ) ?

[1] https://github.com/dmwm/CRABClient/issues/4390#issuecomment-197419726

PR #4611, opened 2016-04-12
Remove code for backward compatibility with old tasks.

AndresTanasijczuk:
The lines being removed were needed for backward compatibility with tasks submitted before CRAB 3.3.1603 (the release with the new crab report code). After 1 month, I guess you don't care anymore about this backward compatibility.


belforte:
thanks for remembering Andres !

PR #4612, opened 2016-04-13
Explicitly test parameter set input configuration for PrivateMC.

### LINKED ISSUE #4610: Cross check generator settings with the pset input source.
We've got some users mixing up `PrivateMC` and the generator settings. It would be best if we (a) cross-check the generator setting with the input source used in the pset or (b) set the generator automatically. I think this is how we could sum up the current possibilites:

* `PrivateMC` + `generator = 'lhe'`: needs to have an `LHESource` in the pset
* `PrivateMC` + `generator = 'pythia'`/default: should have `EmptySource` in the pset
* Everything else really is `Analysis`.

I'll look into how complex this will be.

matz-e:
We receive a lot of emails with misconfigurations when generating MC.  This
will test the input source of the parameter set and try to point the user
to potential fixes before the task hits the crab server and fails.

Fixes #4610.

### DIFF
--- a/bin/crab3bootstrap
+++ b/bin/crab3bootstrap
@@ -59,8 +59,10 @@ def saveCfgInfo(destFile, cmsswCfg):
         info = {}
         edmfiles, tfiles = cmsswCfg.outputFiles()
         lhe, nfiles = cmsswCfg.hasLHESource()
+        pool = cmsswCfg.hasPoolSource()
         info['outfiles'] = [ edmfiles, tfiles]
         info['lheinfo'] = [lhe, nfiles]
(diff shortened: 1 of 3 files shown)


belforte:
@matz-e I read it and can't understand the logic. But for long term maintenance it is better if code passes this test. Maybe there is  a bug, or maybe it is simply a matter of documentation. Can you explain  what you want to do ? even in comments. or IM/voice.. I am at CERN for a few more hours

belforte:
oopps.. I missed that this file is only use for PrivateMC, things make sense now. Sorry. Please @mmascher merge, then I may add a couple inline comments to remind me.

PR #4614, opened 2016-04-22
improve message for #4612

belforte:
made text a bit more clear to reader and to users. @matz-e do you like it ?


matz-e:
Sorry, was traveling. @belforte, thank you, I think should be a bit clearer to the user!

PR #4616, opened 2016-05-02
Fix the crab source script file name

emaszs:


etc/init-light.csh
```diff
@@ -1,2 +1,2 @@
 #!/bin/csh
-eval `sh /cvmfs/cms.cern.ch/crab3/init-light.sh -csh`
+eval `sh /cvmfs/cms.cern.ch/crab3/crab.sh -csh`
```
mmascher: shouldn't it be `crab_light.sh` ?

PR #4628, opened 2016-06-24
Client prototype of the new status

emaszs:
Basic client-side status implementation prototype, has the regular (short) status and the --long status option. It gets the needed summary information directly from the webdir (provided by the task_process daemon) and prints it. Status2 should provide the same information as the old status (in terms of short and long status for the moment), which is what we are planning to test using this new prototype and the old status running in parallel. 

Also needs a PR to be merged on CRABServer: https://github.com/dmwm/CRABServer/pull/5239


src/python/CRABClient/UserUtilities.py
```diff
@@ -132,7 +132,15 @@ def getFileFromURL(url, filename = None, proxyfilename = None):
         socket = opener.open(url)
         status = socket.getcode()
         #TODO Emilis will optimize and read by chunks
```
mmascher: remove the todo and say in a comment why you are doing it in a loop (memory optimization)

src/python/CRABClient/UserUtilities.py
```diff
@@ -149,8 +157,8 @@ def getFileFromURL(url, filename = None, proxyfilename = None):
         exc = ClientException("Unable to retieve the file from %s. HTTP status code %s. HTTP content: %s" % (url, status, socket.info()))
         exc.status = status
         raise exc
-    with open(filename, 'w') as f:
-        f.write(filestr)
+#    with open(filename, 'w') as f:
+#        f.write(filestr)
```
mmascher: Delete?

emaszs:
Removed the useless comment.

Issue #4630, opened 2016-07-05
not handled exception 


belforte:
in
https://github.com/dmwm/CRABClient/blob/master/src/python/CRABClient/JobType/Analysis.py#L181
you should not only assign msg and go on, but likely you want to log the msg and then need to tale some action (raise again or... whatever).
It happened to me that the previous `debugFilesUploadResult = dtb.upload(filecacheurl = filecacheurl)` fails (maybe I have not found the proper combination to make new client work in my private test area...) but then client dies with:
```
UnboundLocalError: local variable 'debugFilesUploadResult' referenced before assignment
```
at next line

separately I (we?) have to investigate why the upload fails. I have updated code in the server and restarted crabcache and crabserver, have updated Oracle.. yet:
```
The server answered with an error.
Server answered with: Input file hashkey mismatch
Reason is: Checksums do not match
```

emaszs:
I don't think raising anything else is necessary if the upload fails - the client and server should be fine without it. The monitor would not fully work in that case, but that's fine. I'll make a PR for handling the exception properly.

belforte:
worth a test, maybe I did something wrong, but I think the server will
complain if it does not find the debug sandbox (at least I got an error
which I attributed to that)

On 07/06/2016 10:32 AM, Emilis Rupeika wrote:
> I don't think raising anything else is necessary if the upload fails - the client and server should
> be fine without it. The monitor would not fully work in that case, but that's fine. I'll make a PR
> for handling the exception properly.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/dmwm/CRABClient/issues/4630#issuecomment-230711127>, or mute the thread
> <https://github.com/notifications/unsubscribe/ABwK2Q_QtFSX4_mD69ZsTKQWZHo083Gvks5qS2g5gaJpZM4JFmDE>.
>

-- 
Stefano Belforte - I.N.F.N.     tel   : +39 040 375-6261 (fax: 375-6258)
Area di Ricerca - Padriciano 99 tel mobile: +39 328 010 7327
34012 TRIESTE TS  - Italy       AIM: stefanobelforte

emaszs:
The server checks if the file exists before copying it to the webdir: https://github.com/dmwm/CRABServer/pull/5236/files#diff-3fa1996a22a9c377b41882d6cf77ab1cR198
[...]

PR #4631, opened 2016-07-06
Check if debug tarball upload result variable exists. #4630

emaszs:
A minor bugfix for when a tarball upload fails and a variable is not initialized.


belforte:
why don't you log the message when there's an error uploading ?

emaszs:
My bad, totally forgot about that. Added logging of the exception, the error message and the traceback appear in the logfile now.

PR #4639, opened 2016-08-19
Duplicate getlog/getoutput commands with some new logic

emaszs:
Added duplicate getlog/getoutput commands to utilize the client-side status and avoid querying the server. The client now uses information returned by status2 to decide which files are possible to retrieve and calculates the pfns by querying phedex. Previously this was done on the crabserver.


mmascher:
In getlog2 I see a lot of copy/paste from getlog (the short log logic). In general I think it is really bad, maybe we can be a bit more relaxed here since getlog is going away.

PR #4649, opened 2016-10-28
Fix 4648

### LINKED ISSUE #4648: make user send the CS UI task URL when they ask for help
Currently users send the dashboard URL, or (more rarely) the uploadlog URL, hardly a task name.
I'd rather have them always point to the CS UI, from where dashboard, config, webdir, log are all easily available (and all in all is what Emilis developed it for).
Before I start advertising for this, I thought it will be good to print this URL to user screen.

@mmascher @emaszs @dciangot what do you think of proposal in following screenshot ?

screenshot from 2016-10-28 11 15 43

belforte:
print CS UI URL - fix #4648 
could not create a branch from master because currenlty it does not run so I could not test it.
Master is broken by report2 importing DBSReader from WMCore. Maybe needs a more fancy setup to test, but I'd prefer if it has less dependencies.


emaszs:
FYI, master client should work fine if wmcore is in the pythonpath. I've always had it in mine, though it doesn't say so in the twiki. Will update it.

PR #4653, opened 2016-12-01
Fix #4648

### LINKED ISSUE #4648: make user send the CS UI task URL when they ask for help

belforte:


src/python/CRABClient/Commands/status.py
```diff
@@ -121,10 +121,17 @@ def printShort(self, dictresult, username):
...
 
-        ## Show dashboard URL for the task.
+        ## Show server and dashboard URL for the task.
+
+        taskname = urllib.quote(self.cachedinfo['RequestName'])
+        ## CRAB Server UI URL for this task is always useful
+        crabServerUIURL = "https://cmsweb.cern.ch/crabserver/ui/task/" + taskname
+        msg = "%sTask URL to use for HELP:\t%s%s" % (colors.GREEN, crabServerUIURL, colors.NORMAL)
+
+	self.logger.info(msg)
```
emaszs: This needs an extra space for correct indentation

src/python/CRABClient/Commands/status2.py
```diff
@@ -163,10 +163,16 @@ def printTaskInfo(self, crabDBInfo, username):
                 msg += "%s" % (status)
         self.logger.info(msg)
 
-        # Show dashboard URL for the task.
+        # Show server and dashboard URL for the task.
+	taskname = urllib.quote(self.cachedinfo['RequestName'])
+
+	## CRAB Server UI URL for this task is always useful
+        crabServerUIURL = "https://cmsweb.cern.ch/crabserver/ui/task/" + taskname
```
emaszs: This has one space too many

belforte:
I do not like tabs either, indeed took a wihile to get the proper number of them :-( But I used them because the nearby code does, so I thought consistency was useful. I know it is not your code, it is not mine either. But changing all prints in this to use spaces may be too much work. Maybe we only do it in status2 ?

emaszs:
What editor are you using? In status.py, I only see two tabs, one in line 131 and another one in line 338, the second one was there for a while so I guess it's fine. In status2.py, I only see 5 tabs coming from your commit on lines 167-174.

emaszs:
Ah, a misunderstanding, I meant the tabs used to indent code, not the tabs used in print statements (these are fine of course). Tabs used to indent code can cause some truly mysterious errors.

belforte:
ah.. right. Sorry. I had started asking if you meant indentation in the code or in the status output in reply to the deleted comment.. then got sidetracked.
Sure, better use spaces in python. I'll fix this. Thanks for spotting.

PR #4663, opened 2017-01-20
Change printout when cores!=threads

emaszs:


mmascher:
Discussed some changes over chat with emilis

PR #4669, opened 2017-01-31
print tarball file in upload message in crab.log Fix #4666

### LINKED ISSUE #4666: print tarball size in log before uploading
add it before this line
https://github.com/dmwm/CRABClient/blob/master/src/python/CRABClient/JobType/Analysis.py#L151
size limit is only enforced in server, but at times uploading may fail due to size and currently there is not telling message. Printing the size in the log will at least allow experts to pin the problem immediately. See  e.g.
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/2560/2.html

belforte:


belforte:
this fixes #4666 

I add this comment since it looks like putting that in the PR title was not effective in the way of having the issue automatically closed when the PR is merged..  github documentation says to put the magic fix keyerong in the BODY of the PR. see
https://github.com/blog/1506-closing-issues-via-pull-requests

emaszs:
@belforte could you add `from __future__ import division` to get the floating point division by default?

emaszs:
Closed accidentally, oops.

belforte:
hmm.. I intentionally wanted sometimes floating point division, sometimes integer. I surely add the import, but.. should I change something else as well ?

emaszs:
That's just to make sure that the code is python 2.7 compatible. You'll need to change your code a bit, `a // b` will be integer division and `a / b` will be floating point division.

belforte:
OK. Understood. I committed the change

src/python/CRABClient/JobType/UserTarball.py
```diff
@@ -9,6 +9,8 @@
 import tarfile
 import tempfile
 
+from  __future__ import division
```
emaszs: Very minor, but could you please move this above the other imports? Below the comment block, that's where we usually put these future imports.

src/python/CRABClient/JobType/UserTarball.py
```diff
@@ -140,13 +142,13 @@ def upload(self, filecacheurl=None):
 
         self.close()
         archiveName = self.tarfile.name
-        archiveSizeKB = os.path.getsize(archiveName)/1024
+        archiveSizeKB = os.path.getsize(archiveName)//1024
         if archiveSizeKB <= 512 :
           archiveSize = "%d KB" % archiveSizeKB
         elif archiveSizeKB < 1024*10 :
           archiveSize = "%3f.1 MB" % (archiveSizeKB/1024.)
```
emaszs: Another very minor thing, the dot after 1024 is no longer necessary for this division to give a floating point result, could be confusing for someone in the future.

src/python/CRABClient/JobType/UserTarball.py
belforte: yes, I am moving the import

src/python/CRABClient/JobType/UserTarball.py
belforte: I understand, but I fear more the danger of confusing me in the short term, if in the distant future someone who only knows python3 will wonder why someone felt the need to make it explicit that "/" indicates floating point division.. well they can clean up at that time !  I.e. I could remove the dot, but then put a comment, would that be better ?

belforte:
branch updated. import __future__ at the top, dot removed, explanatory comments added

emaszs:
Thanks!

PR #4670, opened 2017-02-02
Fix purge command

emaszs:


belforte:
crab purge was working. What was there to fix ?

emaszs:
It's working as far as the cache is concerned, but throws an error about the schedd address you reported here https://github.com/dmwm/CRABServer/issues/5332 and fails to clean up in the schedd. Can you actually get it to work without any errors?

belforte:
OK. No, I do not have a fix for those, was simply confused. So you fixed the error about schedd 
address but not the purge in the schedd, right ? Too bad that the PR did not have a ref. to the 
issue :-(

emaszs:
It should work as intended now, cleaning up the cache and the schedd.

PR #4671, opened 2017-02-07
allow to purge tasks in SUBMITTED status. Fixes #4596

### LINKED ISSUE #4596: Allow to purge the cache area if the state is submitted
It is not possible now: https://github.com/dmwm/CRABClient/blob/master/src/python/CRABClient/Commands/purge.py#L52-L55

belforte:


belforte:
Changes still needed. See:
https://github.com/dmwm/CRABClient/pull/4671#pullrequestreview-22495328

    Only clean the user cache by default.
    Clean up the 'if' statement logic of handling the options, now the ifs are opposite of intuitive.
    If the option to clean the schedd is specified, don't allow tasks in the 'SUBMITTED' status to be purged

PR #4680, opened 2017-03-01
Save downloaded st2 cache file in results dir

emaszs:
Using the getFileFromURL() method the downloaded status_cache file currently gets saved to the directory wherever the command was ran from. With this path, it should be saved to the task's results directory. 

Not sure if we should keep using the getFileFromURL() since there's no real reason in saving the output to a file. I could make a separate method instead, readFileFromURL.


emaszs:
Although the download in getFileFromURL() is optimized to read the file by chunks. The cache file isn't big, but maybe that's enough reason to keep using getFileFromUrl.

PR #4681, opened 2017-03-10
Add --jobids parameter to crab status2 --long and --sort

emaszs:
Issue on CRABServer https://github.com/dmwm/CRABServer/issues/5294

This will print only the selected jobids in crab status --long as well as --sort if the parameter --jobids is used. status --long / status --sort will not be printed if non-existent jobids are given by the user.

### DIFF
--- a/src/python/CRABClient/Commands/status2.py
+++ b/src/python/CRABClient/Commands/status2.py
@@ -9,10 +9,10 @@ from ast import literal_eval
 
 import CRABClient.Emulator
 from CRABClient import __version__
-from CRABClient.ClientUtilities import colors
+from CRABClient.ClientUtilities import colors, validateJobids
 from CRABClient.UserUtilities import getFileFromURL, getColumn
 from CRABClient.Commands.SubCommand import SubCommand
-from CRABClient.ClientExceptions import ClientException
+from CRABClient.ClientExceptions import ClientException, ConfigurationException
 
 from ServerUtilities import TASKDBSTATUSES_TMP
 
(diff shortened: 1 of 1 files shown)


src/python/CRABClient/Commands/status2.py
```diff
@@ -228,7 +245,9 @@ def compareFunction(j1, j2):
             x2 = map(int, j2.split('-'))
             return cmp(x1, x2) #using list comparison
 
-        for jobid in sorted(dictresult.keys(), cmp=compareFunction):
+        # Chose between the jobids passed by the user or all jobids that are in the task
+        jobidsToUse = jobids if jobids else dictresult.keys()
```
mmascher: remove .keys()

src/python/CRABClient/Commands/status2.py
```diff
@@ -92,10 +96,15 @@ def __call__(self):
...
         self.printErrors(statusCacheInfo)
+
         if self.options.summary:
             self.printSummary(statusCacheInfo)
         if self.options.long or self.options.sort:
-            sortdict = self.printLong(statusCacheInfo, quiet = (not self.options.long))
+            # If user correctly passed some jobid CSVs to use in the status --long, self.jobids
+            # will be a list of strings already parsed from the input by the validateOptions()
+            if self.jobids:
+                self.checkUserJobids(statusCacheInfo, self.jobids)
```
mmascher: Should we call it in validate? This is a validation function.

emaszs: It's because we haven't called the DB yet so there's nothing to validate against.

PR #4684, opened 2017-03-20
check tarball size before uploading. Fix #4683

### LINKED ISSUE #4683: check tarball size before uploading and print proper error is larger then limit (100MB atm)
see discussion in comments to #4666

belforte:


src/python/CRABClient/JobType/Analysis.py
```diff
@@ -40,6 +40,7 @@ def run(self, filecacheurl = None):
         Override run() for JobType
         """
         configArguments = {'addoutputfiles'            : [],
+                           'adduserfiles'              : [],
```
emaszs: Hey Stefano, looks like you were using an outdated client branch, this line shouldn't be here, was removed in this commit https://github.com/dmwm/CRABClient/commit/7502a559246128422edf088b57197552f23fc652

src/python/CRABClient/JobType/Analysis.py
```diff
@@ -146,6 +147,7 @@ def run(self, filecacheurl = None):
         with UserTarball(name=tarFilename, logger=self.logger, config=self.config) as tb:
             inputFiles = [re.sub(r'^file:', '', file) for file in getattr(self.config.JobType, 'inputFiles', [])]
             tb.addFiles(userFiles=inputFiles, cfgOutputName=cfgOutputName)
+            configArguments['adduserfiles'] = [os.path.basename(f) for f in inputFiles]
```
emaszs: This line shouldn't be here either.

src/python/CRABClient/JobType/UserTarball.py
```diff
@@ -142,22 +159,29 @@ def upload(self, filecacheurl=None):
...
+        archiveSizeKB = archiveSizeBytes//1024
         if archiveSizeKB <= 512 :
           archiveSize = "%d KB" % archiveSizeKB
         elif archiveSizeKB < 1024*10 :
           # in python3 and python2 with __future__ division, single / means floating point division
           archiveSize = "%3f.1 MB" % (archiveSizeKB/1024)
         else:
           archiveSize = "%d MB" % (archiveSizeKB//1024)
+        if archiveSizeBytes > FILE_SIZE_LIMIT :
+          msg=("%sError%s: input tarball size %s exceeds maximum allowed limit of %d MB" % (colors.RED, colors.NORMAL, archiveSize, FILE_SIZE_LIMIT//1024//1024))
```
mmascher: I think here we should do msg += self.printSortedContent()

src/python/CRABClient/JobType/UserTarball.py
```diff
@@ -142,22 +159,29 @@ def upload(self, filecacheurl=None):
...
+          self.logger.error(msg)
+          raise SandboxTooBigException
```
mmascher: And here SandboxTooBigException(msg) #adding msg

src/python/CRABClient/JobType/Analysis.py
```diff
@@ -158,17 +160,15 @@ def run(self, filecacheurl = None):
...
-                        biggestFileSize = ISBContent[0][0]
-                        ndigits = int(math.ceil(math.log(biggestFileSize+1, 10)))
-                        reason += "\nInput sanbox content sorted by size[Bytes]:"
-                        for (size, name) in ISBContent:
-                            reason += ("\n%" + str(ndigits) + "s\t%s") % (size, name)
+                        reason += " Input sandbox size is ~%sMB. This is bigger than the maximum allowed size of %sMB." % (ISBSize/1024/1024, ISBSizeLimit/1024/1024)
+                        reason += tb.printSortedContent()
                         raise ClientException(reason)
-                raise hte
+            except SandboxTooBigException as e:
```
mmascher: This except is not necessary with the changes proposed below

src/python/CRABClient/JobType/Analysis.py
belforte: rats... I spent so much time with git commands.. and yet this happened. I'll try to patch the PR

src/python/CRABClient/JobType/Analysis.py
belforte: I guess you meant "the changes proposed above".
OK. I will try. I am not really used to exceptions. Guess I have a coding style where they are pretty much handled as exit codes.

belforte:
I have done what @mmascher suggested. But I do not understand why the sandbox content is printed 4 times (sic!) in the crab.log. Maybe there are two loggers, but still ...
I can understand one duplication with two loggers being called in LOGGERS, but am clueless otherwise.
OTOH this is tested and could go in.

belforte:
I have made one last committ where I removed the call to `LOGGERS` (which was there before I edited the code). Now the traceback is only printed once. The exception message with the dump of sandbox content is printed twice in log, which I do not like since it can be  a very long printout, but I suspect this is due to the general exceptoin handling in main `crab`  script where possibly people did not expect a long exception message.
@mmascher in light of this, do you still want to have the sandbox content printout as part of the exception message ?
@mmascher overall i find this `LOGGERS` thing not-documented, confusing, little used and possibly wrongly used as well. Do you think we really need it ? I do not see how future developers/maintainers can find the correct way to use it.

src/python/CRABClient/JobType/Analysis.py
```diff
@@ -158,19 +158,12 @@ def run(self, filecacheurl = None):
                         ISBSize = int(re_match.group(1))
                         ISBSizeLimit = int(re_match.group(2))
                         reason  = "%sError%s:" % (colors.RED, colors.NORMAL)
-                        reason += " Input sanbox size is ~%sMB. This is bigger than the maximum allowed size of %sMB." % (ISBSize/1024/1024, ISBSizeLimit/1024/1024)
-                        ISBContent = sorted(tb.content, reverse=True)
-                        biggestFileSize = ISBContent[0][0]
-                        ndigits = int(math.ceil(math.log(biggestFileSize+1, 10)))
```
emaszs: Sorry @belforte, one other minor thing, could you remove the unused `math` import?

belforte: Done.

mmascher:
> in light of this, do you still want to have the sandbox content printout as part of the exception message ?

That's what I was thinking, instead of

belforte:
@mmascher to be clear I could not understand why the sandbox content (i.e. the exception message) is printed twice. I am sure we can live with current double printout, it is mostly for my education, whatever that may be worth. If you know how to solve, tell me, or just do it.  If I had time to cleanup the client log, I'd rather try to get rid of all those messages from proxy handling in WMCore.

mmascher:
It is because of those two lines: https://github.com/dmwm/CRABClient/blob/master/bin/crab#L197-L198

```python
import logging
try:
   raise Exception("ciao")
except Exception as ex:
  logging.error(ex)
  logging.exception("Caught exception")
```

will print

```
ERROR:root:ciao
ERROR:root:Caught exception
  File "<stdin>", line 2, in <module>
Exception: ciao
```

emaszs:
Wait, so do we need the first logging call if it only contains the exception message? The second one should also have the message plus the exception?

mmascher:
Yes, we need both: .error is what the user sees (The error message from ClientExceptions is something the user want to see), while .exception prints the stacktrace for the operators. I'd like to keep the stacktrace honestly even though we duplicate the error message, sometimes it is useful to track down exactly when an error happened.

belforte:
in other words: we intentionally print all exception messages twice in the log and too bad that in some particular case like this it can be a very long message. Indeed when message is one line, there is no problem in having it replicated after the traceback.

mmascher:
yes. Now that I think about it, I guess we can use something else than the .exception method if this is annoying. Probably print_stack here should do it: https://docs.python.org/2/library/traceback.html

PR #4688, opened 2017-04-12
Update bash completion script

matz-e:
See https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/2781.html

I just ran `scripts/generate_completion.py`…

### DIFF
--- a/etc/crab-bash-completion.sh
+++ b/etc/crab-bash-completion.sh
@@ -23,13 +23,13 @@ _UseCrab ()
         "")
             case "$cur" in
             "")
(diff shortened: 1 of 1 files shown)


matz-e:
Uhm… @ericvaandering that bot behavior does not seem right to me…

ericvaandering:
Yeah, the CRAB bot has been useless for a while now. On my list of things to re-vamp

emaszs:
Thank you @matz-e. @belforte, @mmascher, should I merge this and deploy a patch release for the production client? I can trust that this should work but I'm not too familiar, also I won't be around next week. Since I haven't started switching the client commands yet the autocompletion will need to be generated again during validation.

belforte:
do we still have a pre-prod client ? If so, can you update that, but leave production as is ?
Then worry about production client once you are back.

emaszs:
Yes of course, I'll apply this to pre-prod then.

PR #4689, opened 2017-04-24
Rename commands

emaszs:
I've renamed:
getlog -> getlogold
getoutput -> getoutputold
report -> reportold
resubmit -> resubmitold
status -> statusold

and 
getlog2 -> getlog
getoutput2 -> getoutput
report2 -> report
resubmit2 -> resubmit
status2 -> status

So far I haven't touched the server, there's various API names that need renaming also (like logs, logs2 in HTCondorDataWorkflow) but I think they can wait until we start removing the old commands.


belforte:
@emaszs is this understanding of mine correct then ?
1. this PR  client so that `crab status` calls 'crab status2` on the server and 'crab statusold' calls 'crab status' on the server etc.
2. at some later point you remove `*old` commands from client.
3. after that you can modify the server so that `status` does the same as `status2` etc,
4. after that you can modify the client so that `status` calls `status` on server (no use for `status2` anymore)
5. finally remove `status` etc. from server.

Or do you have a different plan in mind ?

emaszs:
Oops, didn't see your comment before merging. Yep, I'm thinking the same, we'll get rid of the old client commands and have two placeholders on the server doing the same new thing before removing 'status2' references on the server and redirecting the client queries to the final correct APIs.

There's no status2 on the server, it's report2, resubmit2, output2/logs2 which will need to be modified in the future to reflect the change on the client. But yes, for now `crab report` will call report2 on the server.

PR #4694, opened 2017-05-03
Check if status2 succeeded before executing getcommand

emaszs:
This should handle the case when the status information is unavailable for whatever reason and we cannot execute the rest of getcommand. Without this an error is printed to stdout: `TypeError: 'NoneType' object has no attribute '__getitem__'`.


emaszs:
Actually I had already fixed it along with the API improvements

PR #4695, opened 2017-05-04
Use proxied webdir for file downloads in new commands

emaszs:
Not exactly well tested but should work. I had to refactor some other code, in crab report I now use getFileFromUrl instead of constructing a pycurl request because I think it makes it easier to pass the user's proxy with the request.

### DIFF
--- a/src/python/CRABClient/Commands/report.py
+++ b/src/python/CRABClient/Commands/report.py
@@ -2,9 +2,7 @@ from __future__ import print_function, division
 
 import os
 import json
-import tempfile
 import pycurl
-import StringIO
 import tarfile
 
 import CRABClient.Emulator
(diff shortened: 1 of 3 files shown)


belforte:
no need to hurry. Can take the time to test well.

emaszs:
While I wait for Marco to review I'll deploy this client to preprod

src/python/CRABClient/Commands/report.py
```diff
@@ -351,37 +348,26 @@ def getLumisToProcess(self, userWebDirURL, numJobs, workflow):
...
-                                    fd.close()
-                    finally:
-                        tarball.close()
-            finally:
-                fp.close()
-                hbuf.close()
+                                res[str(jobid)] = json.load(fd)
+                            finally:
+                                fd.close()
+            except ClientException as ce:
```
mmascher: One small detail, if you pick `LOGGERS['CRAB3']` from https://github.com/dmwm/CRABClient/blob/master/src/python/CRABClient/ClientUtilities.py#L96 it will be possible to print the exception traceback only in the logfile. It might be useful imho.

Issue #4698, opened 2017-05-08
CRAB Client API backward compatibility broken


belforte:
Apparently after last release, crab API returns a ntuple instead of a dictionary[1].
This broke HammerCloud and I am puzzled that other users have not complained yet.
HammerClould people had to patch their code in a hurry because all tests stopped working suddenly. That's clearly something unacceptable.
Do we use client API in the validation ?

[1]
```
>>> statusresult = crabCommand('status', dir = crab_work_dir)
>>> type(statusresult)
<type 'tuple'>
>>> type(statusresult[0])
<type 'dict'>
>>> len(statusresult)
2
>>> type(statusresult[1])
<type 'dict'>
>>> statusresult[1]
{'jobList': [['finished', '1'], ['finished', '2']], 'jobsPerStatus': {'finished': 2}}
>>> statusresult[0]
{'status': u'SUBMITTED', 'username': u'belforte', 'jobs': {'1': {'State': 'finished'}, '2': {'State': 'finished'}}, 'jobsPerStatus': {'finished': 2}, 'taskWarningMsg': u'[]', 'schedd': u'crab3@vocms0106.cern.ch', 'publicationFailures': {}, 'collector': u'cmsgwms-collector-global.cern.ch:9620,cmssrv221.fnal.gov:9620', 'proxiedWebDir': u'https://cmsweb.cern.ch/scheddmon/0106/cms1627/170424_173550:belforte_crab_20170424_193545', 'statusFailureMsg': '', 'taskFailureMsg': None, 'inputDataset': u'/GenericTTbar/HC-CMSSW_7_0_4_START70_V7-v1/GEN-SIM-RECO', 'userWebDirURL': u'http://vocms0106.cern.ch/mon/cms1627/170424_173550:belforte_crab_20170424_193545', 'ASOURL': u'https://cmsweb.cern.ch/crabserver/prod', 'publication': {'disabled': []}, 'submissionTime': 1493055350, 'command': u'SUBMIT', 'outdatasets': u"['/GenericTTbar/belforte-CRAB3_test-SB-bb695911428445ed11a1006c9940df69/USER']", 'publicationEnabled': False}
>>> 

```

belforte:
to be clear, we need to answer the usual questions here (urgently):
A) do we need to remedy ? how ? i.e.
  1. will anybody else be affected ?
  2. if yes to above:  how serious and what can we do ?
B) what do we have to do so that it does not happen again ?

belforte:
[...]

PR #4699, opened 2017-05-08
Make new status calls return only one value

emaszs:


emaszs:
https://github.com/dmwm/CRABClient/issues/4698

Issue #4702, opened 2017-05-08
review how status is reported in API


belforte:
see
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/2859/1/1.html
the new status API reports 
```
{'status': u'SUBMITTED', 'username': u'tpook', ....
```
where the old one would have written
```
{u'status': u'FAILED', u'username': u'tpook' ....
```
i.e. new API reports as `status` the status on the server, not the DAG one.

Discussing with @emaszs he pointed out that old API creates a combinatino of task and dag status using a somehow complex logic.
Hopefully most of what was done in there, is now obsolete since we have task process, but we need to review this to make sure that we do not have gross, obvious inconsistencies like the one above.
In ubtle cases hopefully we simply do better now.

belforte:
for reference, I paste here pointers to old implementation from the chat with @emaszs 
in the old version, this is the task status that gets filled in in this method and later returned to the client https://github.com/dmwm/CRABServer/blob/master/src/python/CRABInterface/HTCondorDataWorkflow.py#L459
the failures affect the task status https://github.com/dmwm/CRABServer/blob/master/src/python/CRABInterface/HTCondorDataWorkflow.py#L510-L527
 task status for finished tasks https://github.com/dmwm/CRABServer/blob/master/src/python/CRABInterface/HTCondorDataWorkflow.py#L580-L588
also https://github.com/dmwm/CRABServer/blob/master/src/python/CRABInterface/HTCondorDataWorkflow.py#L692-L693

belforte:
I had a chat with @emaszs on this topic this morning.
Here's the executive  summary. @emaszs please review that I remembered correctly and if you agree, can you provide a time estimate for implementation ?

1. keep reporting ony one `'status'`key in the dictionary
2. until task status in database reaches SUBMITTED, report the status in DB as task status
3. when task is SUBMITTED, check if webdir info was put in DB by the dagbootstrap
4. if answer to 3 is yes, lookup status_cache file created by TaskProcess and report that
5. if answer to 3. is no, use a condor_query to check if dag bootstrap job is pending/running or crashed
6. after 5. if dag bootstrap is pending/running, report SUBMITTED
7. after 5. if dag bootstrap is crashed, report FAILED
8. if at any time can not talk to server, or to schedd, report UNKNOWN
9. it is up to the user to cache previsous status and use it in place of the new UNKNOWN
[...]

PR #4710, opened 2017-05-26
Combine task statuses into a single one

emaszs:
According to the logic outlined here: https://github.com/dmwm/CRABClient/issues/4702#issuecomment-303737455. I'll review next week with @mmascher when he's around though now I'm pretty sure this will do what we said it would in that thread.

### DIFF
--- a/src/python/CRABClient/Commands/status.py
+++ b/src/python/CRABClient/Commands/status.py
@@ -7,6 +7,7 @@ import urllib
 import logging
 from ast import literal_eval
 from datetime import datetime
+from httplib import HTTPException
 
 import CRABClient.Emulator
 from CRABClient import __version__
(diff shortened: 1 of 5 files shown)


src/python/CRABClient/Commands/status.py
```diff
@@ -57,7 +60,7 @@ def __call__(self):
         if not rootDagId:
             failureMsg = "The task has not been submitted to the Grid scheduler yet. Not printing job information."
```
belforte: I would not call this a failure. In such cases DB status is NEW or QUEUED. Only if DB status is SUBMITTED and the rootDagId missing we can call it an error

emaszs: This is shorthand for statusFailureMsg and it is a failure in the sense that the status cannot provide all of the information it normally would.

belforte: please see https://github.com/dmwm/CRABClient/issues/4702#issuecomment-304273801
We should not depend on explanations in comments here in order to argue on the meaning of things.

PR #4717, opened 2017-07-31
Add client-side check for max number of output file lumis. Fix #4706

### LINKED ISSUE #4706: prevent users from creating pprivateMC files with too many empty lumis
see also:
https://github.com/dmwm/CRABServer/issues/5488
https://github.com/dmwm/CRABServer/issues/5490

when user do PrivateMC they specify how many events crab generates/simulates, but
they may have filters so only a very small fraction goes in output file.

We do not want those files to be filled with empty lumis.

Currently user speicfy the desired # of genrated events as Unit in crab config. And they set
```
unitsPerJob
totalUnits

```
while this is usually left at its default value
```
eventsPerLumi = 100
```

we should compute the number of lumis per file as `lumisPerFile=unitsPerJob/eventsPerLumi` and declare error if `lumisPerFile>1k` or something similar

emaszs:
As @belforte suggested in https://github.com/dmwm/CRABClient/issues/4706, this is a simple pre-submission check on how many lumis the user MC output files will have based on their configuration parameters. 

Though now re-reading the issue, do I understand it correctly? Checking the configuration params doesn't hurt but AFAIU that's not really how problematic datasets are produced. Instead, users generate events and then discard some of them before writing to output, so the "real" checking should happen before publication.

### DIFF
--- a/src/python/CRABClient/JobType/PrivateMC.py
+++ b/src/python/CRABClient/JobType/PrivateMC.py
@@ -1,8 +1,11 @@
 """
 PrivateMC job type plug-in
 """
+from __future__ import division
+from __future__ import print_function
 
 import os
+import math
 
 from CRABClient.ClientUtilities import colors
 from CRABClient.JobType.Analysis import Analysis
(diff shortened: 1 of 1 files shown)


ericvaandering:
test this please

belforte:
@ericvaandering in case you did not know, the Details link for the python3 compatibility test gives error.

ericvaandering:
Yup. This was the first attempt. There are a couple of problems (that's one).

ericvaandering:
test this please

ericvaandering:
test this please

ericvaandering:
test this please

ericvaandering:
test this please

ericvaandering:
test this please

ericvaandering:
test this please

ericvaandering:
Ok, finally got a good run of the client tests.

src/python/CRABClient/JobType/PrivateMC.py
```diff
@@ -31,7 +31,7 @@ def run(self, *args, **kwargs):
             self.logger.debug("LHESource found in the CMSSW configuration.")
             configArguments['generator'] = getattr(self.config.JobType, 'generator', 'lhe')
 
-            major, minor, fix = os.environ['CMSSW_VERSION'].split('_')[1:4]
+            major, minor, _ = os.environ['CMSSW_VERSION'].split('_')[1:4]
```
belforte: does it really make sense to use `_` as a variable ? I am surely one person who gets heavily confused by this. Is that some kind of PEP-8 standard for a variable which is never utilized ?

belforte: OK. I found the answer by myself, sorry for noise. `_` is indeed Python conding convention for throwaway variable. I learnt one more thing.

emaszs: We can use something like `_dummy`, `_dummy_fix` or `_unused`, `_unused_fix` etc, but I think the code here simple enough without the need of any extra explanations.

belforte:
@mmascher please merge.

This PR suffers of a longish comment history unrelated to the topic. So in the CRAB meeting it was felt the need to clarify where we are.
I think it is fine and can be merged. About initial questions from @emaszs in https://github.com/dmwm/CRABClient/pull/4717#issue-246768812 : the event filtering done by users may affect how large is the output file, but the number of lumis is independent, empy lumis are still listed in the output, furthermore the number of events in the configuration (checked in the code that @emaszs provided) is the number of events to be generated and simulated, not to be output. If users filter heavily, they will tend to ask for a large number of events in each file in order to get some decent statistics in output, and that's the very reason to make sure that those files do not have too many lumis.

PR #4720, opened 2017-08-10
Fixes for automatic splitting

matz-e:
Skip probe jobs where appropriate.


matz-e:
I've tested this now with tailjobs, @emaszs, and it looks like it's working. I've moved loading the cached configuration file earlier, to make validating the passed jobIds easier.

PR #4741, opened 2018-02-01
Adding the etc/init-light-pre.sh script && Changing the sequence of e…

todor-ivanov:
…ditting the PYTHONPATH environment variable.


mmascher:
Looks good to me. I'll let @belforte have a look, the point of this is to avoid having to set the CRAB_SOURCE_SCRIPT for using the preprod client, which is confusing. Todor and I agreed this is a good solution.

belforte:
LGTM

PR #4738, opened 2018-01-25
Improve automatic splitting et al.

matz-e:
Companion to dmwm/CRABServer#5627:

* Change units for automatic splitting to minutes rather than seconds
* Make automatic splitting default splitting algorithm
* Accept fractional `Data.totalUnits`
* Improve automatic splitting status:
  * Don't display splitting and failed jobs by default in summary
  * Only split job display into probes, processing, tails with `--long`
  * Mark failed processing jobs as `rescheduled`
  * Probe jobs are `no output` rather than `failed`/`finished`
  * Don't allow to resubmit processing jobs with automatic splitting

PR #4746, opened 2018-02-08
Make sure that "get" commands accept default splitting.

matz-e:
Fix getoutput command as discovered by @lecriste in #4738.


matz-e:
@lecriste this fix works for me. Could you confirm?

@belforte for the client to be fully usable with pure default settings (no `splitting` parameters at all when running in analysis mode), this will make the client fully usable.

lecriste:
I confirm this fixes the last [problem](https://github.com/dmwm/CRABClient/pull/4738#issuecomment-363828095).

Issue #4739, opened 2018-01-28
crab status --long


lecriste:
- The jobs "Most Recent Site" column shows always "Unknown"
- Print the Summary also in the plain `crab status` command

lecriste:
This fix the bug in the "`Most Recent Site`" column:
https://gitlab.cern.ch/ai/it-puppet-hostgroup-vocmsglidein/commit/7e69fc91c015ac0f6ae1a25aeb3d36807d6f011b

lecriste:
Fixed by #4748

Issue #4732, opened 2018-01-18
fix crab status --summary


belforte:
currently crab status --summary output is empty (at least in some cases) [1]
shoudl at least have the gransummary printed at the end of crab status --long:
```
Summary:
 * Memory: 1232MB min, 1464MB max, 1251MB ave
 * Runtime: 2:59:11 min, 12:32:49 max, 5:58:04 ave
 * CPU eff: 67% min, 98% max, 95% ave
 * Waste: 27:12:38 (3% of total)
```
[1]
```
belforte@lxplus099/sal> crab status -d crab_JetHT_Run2016D-07Aug17-v1/ --summary 
Warning: Incompatible CRABClient version development
Server is saying that compatible versions are: [u'3.3.1704', u'3.3.1705', u'3.3.1706', u'3.3.1707', u'3.3.1711', u'3.3.1711.rc4']
CRAB project directory:		/afs/cern.ch/work/b/belforte/CRAB3/TC3/dbg/sal/crab_JetHT_Run2016D-07Aug17-v1
Task name:			180115_161640:srappocc_crab_JetHT_Run2016D-07Aug17-v1
Grid scheduler:			crab3@vocms0195.cern.ch
Status on the CRAB server:	SUBMITTED
Task URL to use for HELP:	https://cmsweb.cern.ch/crabserver/ui/task/180115_161640%3Asrappocc_crab_JetHT_Run2016D-07Aug17-v1
Dashboard monitoring URL:	http://dashb-cms-job.cern.ch/dashboard/templates/task-analysis/#user=srappocc&refresh=0&table=Jobs&p=1&records=25&activemenu=2&status=&site=&tid=180115_161640%3Asrappocc_crab_JetHT_Run2016D-07Aug17-v1
Status on the scheduler:	COMPLETED

Jobs status:                    finished      100.0% (146/146)

Publication status:		done           93.8% (137/146)
				failed          6.2% (  9/146)

Output dataset:			/JetHT/srappocc-JetHT_Run2016D-07Aug17-v1-b84139e74cafd5a0bfdaf4226ec7e7b4/USER
Output dataset DAS URL:		https://cmsweb.cern.ch/das/request?input=%2FJetHT%2Fsrappocc-JetHT_Run2016D-07Aug17-v1-b84139e74cafd5a0bfdaf4226ec7e7b4%2FUSER&instance=prod%2Fphys03

Site Summary Table (including retries):

[...]

PR #4748, opened 2018-02-19
Status issues

lecriste:
Warn user of poor resources usage and fix `crab status --summary`.


src/python/CRABClient/Commands/status.py
```diff
@@ -355,7 +359,7 @@ def checkUserJobids(self, statusCacheInfo, userJobids):
         if wrongJobIds:
             raise ConfigurationException("The following jobids were not found in the task: %s" % wrongJobIds)
 
-    def printDetails(self, dictresult, jobids=None, quiet=False):
+    def printDetails(self, dictresult, jobids=None, quiet=False, maxMemory=2000, maxJobRuntime=1315):
```
belforte: default valules should not be hardwired in two places. Need to pick maxMemory and maxJobRuntime via e.g.
`from CRABClient.ClientMapping import parametersMapping, getParamDefaultValue`

belforte:
I merged to facilitate testing, but this PR needs a bit of work.
1)
See comment:
https://github.com/dmwm/CRABClient/pull/4748#pullrequestreview-97787394
2)
what exactly is the issue with site history ? Looks like you are trying to protect the code against
using an empty list, but the problem is that site must be known for a job which ran.

belforte:
@lecriste in the commit comment please refer to which issue you are fixing, so that github will close automatically when the commit is merged in the master (besides having the cross refs from code to issue)

lecriste:
I agree with 1), wanted to do that way already but did not find the functions, thanks for pointing me.
2) The Unknown site problem was due to a [misconfiguration](https://github.com/dmwm/CRABClient/issues/4739#issuecomment-366291526) in the puppet repo. The protections I added fix this [issue](https://github.com/dmwm/CRABClient/issues/4732) in case the site is unknown for any reason and help the debugging.

PR #4750, opened 2018-03-22
Resolve #4740

### LINKED ISSUE #4740: in crab status output make it clear where Publication info is from
In current status output make it print that Publication status refers to information from output dataset in DBS
```
Jobs status:                    finished      100.0% (481/481)

Publication status:		done           98.5% (474/481)
				new             1.5% (  7/481)
```

lecriste:
The client queries the RESTInterface which executes
https://github.com/dmwm/CRABServer/blob/master/src/python/CRABInterface/HTCondorDataWorkflow.py#L933

with this Oracle FileTransfersDB query:
https://github.com/dmwm/CRABServer/blob/9b3837b94aeb26fde76790c111e13f4f2034a90d/src/python/Databases/FileTransfersDB/Oracle/FileTransfers/FileTransfers.py#L246

### DIFF
--- a/src/python/CRABClient/Commands/status.py
+++ b/src/python/CRABClient/Commands/status.py
@@ -908,7 +908,7 @@ class status(SubCommand):
             self.logger.info(msg)
             return pubStatus
         ## Case in which there was an error in retrieving the publication status.
-        pubSource = "\n(from Oracle transfer DB)"
+        pubSource = "\n(from internal CRAB transfersDB)"
         if 'error' in pubInfo['publication']:
             msg = "\nPublication status:\t\t%s" % (pubInfo['publication']['error'])
             msg += pubSource
--- a/src/python/CRABClient/Commands/status.py
+++ b/src/python/CRABClient/Commands/status.py
@@ -908,8 +908,10 @@ class status(SubCommand):
             self.logger.info(msg)
             return pubStatus
         ## Case in which there was an error in retrieving the publication status.
+        pubSource = "\n(from Oracle transfer DB)"
         if 'error' in pubInfo['publication']:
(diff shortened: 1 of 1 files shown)


src/python/CRABClient/Commands/status.py
```diff
@@ -908,8 +908,10 @@ def printPublication(self, publicationEnabled, jobsPerStatus, numProbes, numUnpu
             self.logger.info(msg)
             return pubStatus
         ## Case in which there was an error in retrieving the publication status.
+        pubSource = "\n(from Oracle transfer DB)"
```
belforte: let's put something whjich makes sense to the user, and yet is useful to us,  like "from internal CRAB transferdb"

belforte: better: "from internal CRAB bookkeeping in transferdb"

src/python/CRABClient/Commands/status.py
```diff
@@ -908,7 +908,7 @@ def printPublication(self, publicationEnabled, jobsPerStatus, numProbes, numUnpu
             self.logger.info(msg)
             return pubStatus
         ## Case in which there was an error in retrieving the publication status.
-        pubSource = "\n(from Oracle transfer DB)"
+        pubSource = "\n(from internal CRAB transfersDB)"
```
belforte: sorry, names matter. The text I propose is better.
"CRAB internal bookkeeping" is something which users can understand
transferdb (no uppercase) is the name of the table inside CRAB ORACLE DB which is used to keep track, i.e. it is the details needed for experts.
There is no "transfer DataBase " in CRAB separate from other DataBase(s). The table name is not ideal, but that was not something I picked and fear it is too much work to put a decent name there.

PR #4754, opened 2018-04-10
Status issues

lecriste:
Do not print warning messages when jobs use less than the default amount of RAM or when no jobs have run yet.

### DIFF
--- a/src/python/CRABClient/Commands/status.py
+++ b/src/python/CRABClient/Commands/status.py
@@ -465,14 +465,14 @@ class status(SubCommand):
         if mem_cnt or run_cnt:
             # Print a summary with memory/cpu usage.
             hint = " and/or improve the jobs splitting (e.g. config.Data.splitting = 'Automatic') in a new task"
-            usage = {'memory':[mem_max,maxMemory,0.7,parametersMapping['on-server']['maxmemory']['default'],'MB'], 'runtime':[to_hms(run_max),maxJobRuntime,0.3,parametersMapping['on-server']['maxjobruntime']['default'],'min']}
+            usage = {'memory':[mem_max,maxMemory,0.7,parametersMapping['on-server']['maxmemory']['default'],'MB'], 'runtime':[to_hms(run_max),maxJobRuntime,0.3,0,'min']}
             for param, values in usage.items():
                 if values[1] > values[3]:
                     if values[0] < values[2]*values[1]:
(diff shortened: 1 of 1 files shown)


src/python/CRABClient/Commands/status.py
```diff
@@ -465,10 +465,11 @@ def translateJobStatus(jobid):
         if mem_cnt or run_cnt:
             # Print a summary with memory/cpu usage.
             hint = " and/or improve the jobs splitting (e.g. config.Data.splitting = 'Automatic') in a new task"
-            usage = {'memory':[mem_max,maxMemory,0.7,'MB'], 'runtime':[to_hms(run_max),maxJobRuntime,0.3,'min']}
+            usage = {'memory':[mem_max,maxMemory,0.7,parametersMapping['on-server']['maxmemory']['default'],'MB'], 'runtime':[to_hms(run_max),maxJobRuntime,0.3,maxJobRuntime=parametersMapping['on-server']['maxjobruntime']['default'],'min']}
```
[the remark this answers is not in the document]
belforte: hmm we should warn if user sticks to default 20h but uses only a few min.  Even if we will be able to time tune, a decent estimate always helps.

src/python/CRABClient/Commands/status.py
```diff
@@ -465,14 +465,16 @@ def translateJobStatus(jobid):
...
-            if cpu_ave < cpu_thr:
-                self.logger.info("\n%sWarning%s: the average jobs CPU efficiency is less than %d%%, please consider to request a lower number of threads%s." % (colors.RED, colors.NORMAL, cpu_thr*100, hint))
+                if values[1] > values[3]:
+                    if values[0] < values[2]*values[1]:
+                        self.logger.info("\n%sWarning%s: the max jobs %s is less than %d%% of the task requested value (%d %s), please consider to request a lower value (allowed through crab resubmit)%s." % (colors.RED, colors.NORMAL, param, values[2]*100, values[1], values[4], hint))
+            if run_sum:
+                cpu_ave = (cpu_sum / run_sum)
+                cpu_thr = 0.7
+                if cpu_ave < cpu_thr:
+                    self.logger.info("\n%sWarning%s: the average jobs CPU efficiency is less than %d%%, please consider to request a lower number of threads%s." % (colors.RED, colors.NORMAL, cpu_thr*100, hint))
```
belforte: @lecriste you should make sure that user has requeste >1 thread before suggesting to lower their number. Maybe we should stick to 50% for such a generic warning, since low CPU eff may come from many sources and users may not have a clue at what to do

lecriste: Ok.
What do we write in the warning when the number of thread is 1?

PR #4756, opened 2018-04-17
Print warnings only with at least one finished job

lecriste:
Avoid getting fake warnings because based on parameters from not finished jobs (e.g. in the early stage of a task).

### DIFF
--- a/src/python/CRABClient/Commands/status.py
+++ b/src/python/CRABClient/Commands/status.py
@@ -398,13 +398,13 @@ class status(SubCommand):
                 return 'rescheduled'
             return statusToTr
 
-        oneJobFinished = False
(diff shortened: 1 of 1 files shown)


belforte:
hmm... seems you are making code more complex . Why not simply to skip jobs which have not run yet from all kind of summary metrics moving up (a possibly slightly modified copy of) this line ?
https://github.com/lecriste/CRABClient/blob/2f715f4cbb0ca0e92e1403cef297780d052623be/src/python/CRABClient/Commands/status.py#L434

Besides it looks to me that the code in this PR will still average RAM usage over idle jobs, simply do not make the average unless at least one job completed. Or am I misreading ?

Issue #4755, opened 2018-04-17
checkwrite and getoutput not working when gfal-copy in non-system path


chwissing:
Hi,
at DESY we provide Grid clients via CVMFS. In order to reach them the path needs to be adjusted properly. It looks like that crab client does not work properly, when gfal-copy is not found in the standard system path like /usr/bin. Find a typical error message below[*]. Appears to be an environment issue, the special path is not properly preserved.

Cheers, Christoph

[*]
rab checkwrite --site=T2_DE_DESY
     Will check write permission in the default location
     /store/user/<username>
     Retrieving DN from proxy...
     DN is: /C=DE/O=GermanGrid/OU=uni-hamburg/CN=XXXXXX
     Retrieving username from SiteDB...
     Username is: XXXXX
     Validating LFN /store/user/XXXXX...
     LFN /store/user/lbenato is valid.
     Will use `gfal-copy`, `gfal-rm` commands for checking write
permissions
     Will check write permission in /store/user/XXXXX on site
T2_DE_DESY

     Attempting to create (dummy) directory
     crab3checkwrite_20180416_131300 and copy (dummy) file
     crab3checkwrite_20180416_131300.tmp to /store/user/XXXXX

     Executing command: env -i X509_USER_PROXY=/tmp/x509up_u30466
     gfal-copy -p -v -t 180
    
file:///afs/desy.de/user/l/lbenato/LongLivedTriggers/CMSSW_8_0_26_patch1/src/Analysis/ALPHA/crab3checkwrite_20180416_131300.tmp

    
'srm://dcache-se-cms.desy.de:8443/srm/managerv2?SFN=/pnfs/desy.de/cms/tier2/store/user/XXXXX/crab3checkwrite_20180416_131300/crab3checkwrite_20180416_131300.tmp'

     Please wait...
     Failed running copy command
       Stderr:
         env: gfal-copy: No such file or directory
       
     Checkwrite Result:
     Unable to check write permission in /store/user/lbenato on site
     T2_DE_DESY

belforte:
Is this something that always happened but was never reported, or should we look for some recent change ? I am puzzled because I thought nothing changed here in years.
The goal in crab client is to execute grid commands in the same env. in which crab command
was issued, does gfal-coipy work before the user types "crab submit" ?
I suspect there's no easy way for me to log in at DESY and check details, right ?
[...]

PR #4758, opened 2018-04-20
Fix 4755

belforte:
do not use env -i before gfal-* but simply undo scram environment as a general protection against CMSSW vs. GRID conflicts
Also  a bit of code cleanup for readibility

### DIFF
--- a/src/python/CRABClient/Commands/remote_copy.py
+++ b/src/python/CRABClient/Commands/remote_copy.py
@@ -78,7 +78,7 @@ class remote_copy(SubCommand):
         command = ""
         if cmd_exist("gfal-copy") and self.options.command not in ["LCG"]:
             self.logger.info("Will use `gfal-copy` command for file transfers")
-            command = "env -i X509_USER_PROXY=%s gfal-copy -v " % os.path.abspath(self.proxyfilename)
+            command = "gfal-copy -v "
             if self.options.checksum:
                 command += "-K %s " % self.options.checksum
             command += " -T "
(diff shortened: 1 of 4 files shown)


src/python/CRABClient/Commands/remote_copy.py
```diff
@@ -140,10 +140,13 @@ def __call__(self):
                 continue
 
             ##### Creating the command
+            # better to execut grid commands in the pre-CMS environment
+            undoScram = "which scram >/dev/null 2>&1 && eval `scram unsetenv -sh`"
+            cmd = undoScram + cmd
```
lecriste: `cmd` is not defined before, any reason to add `+ cmd`?
Anyhow it is overridden by https://github.com/dmwm/CRABClient/pull/4758/files#diff-62f68c1aa854c51b75d7304faaff2af7R153 and not used before.

belforte: thanks @lecriste  for spotting this. It was a plain mistake.  But I do not understand what you write about it being overridden. I do not understand if I should fix this or you did somethig already, or ... ?

lecriste: `cmd` is assigned in L145 and reassigned in L153 without being used before, so the former value is never used.
I did not touch the code.

belforte: thanks, understood. never the less L145 will cause a failure ! I will fix this.

belforte: also the way is is coded now will plainly not work !

belforte:
@lecriste 
this was tested at DESY as well and it solves the problem #4755 . I propose to merge,tag and deploy in preprod toghehter with the new WMCore tag.

lecriste:
+1
Shall I merge and tag or do you want to do it?
To go in preprod the tag shoul be be 3.3.1805.rc2

belforte:
please do it. I would most likely get confused with server and mess things up :-(

PR #4761, opened 2018-05-12
Pointer to Automatic slitting FAQ

lecriste:
Some users asked for clarification because they did not read the documentation.

### DIFF
--- a/src/python/CRABClient/Commands/status.py
+++ b/src/python/CRABClient/Commands/status.py
@@ -505,7 +505,7 @@ class status(SubCommand):
     def printOverview(self, statusCacheInfo):
         """ Give a summary of the job statuses, keeping in mind that:
                 - If there is a job with id 0 then this is the probe job for the estimation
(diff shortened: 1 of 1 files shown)


src/python/CRABClient/Commands/status.py
```diff
@@ -566,6 +566,9 @@ def terminate(states, jobStatus, target='no output'):
         # And if the dictionary is not empty, print it
         for jobtype, currStates in toPrint:
             if currStates:
+                automaticSplittFAQ = 'https://twiki.cern.ch/twiki/bin/view/CMSPublic/CRAB3FAQ#What_is_the_Automatic_splitting'
+                if any('-' in k for k in statusCacheInfo):
```
lecriste: @belforte, this 'any' is used already twice in the code.
Should we rather define a flag in the __call__ and pass it to the different functions?

belforte:
looks good.

PR #4763, opened 2018-05-18
Allow to submit jobs with `numCores = 2`

lecriste:


ericvaandering:
test this please

ericvaandering:
Someone should go and fix that python3 compatibility bug which was introduced somewhere and is going to cause all the tests to fail that part until fixed.

belforte:
@ericvaandering could not agree more, but no me nor @lecriste could figure out where/what to change

the line quoted in the report is already fixed, AFAICS
https://github.com/dmwm/CRABClient/blob/master/src/python/CRABClient/Commands/SubCommand.py#L183

ericvaandering:
Interesting. My guess is that there is no release since that change was made a few months ago, or maybe that the Jenkins container has stale code and that’s the problem?

Important thing is that you fixed it. :-) I can see about Jenkins if you like

belforte:
There's several releases after it was fixed. Could you check Jenkins ? Thanks.

PR #4769, opened 2018-06-08
Avoid usage of '--dryrun' with Automatic splitting

lecriste:


src/python/CRABClient/Commands/submit.py
```diff
@@ -190,6 +190,11 @@ def validateConfig(self):
                 msg = "Invalid CRAB configuration: Parameter General.requestName should not be longer than %d characters." % (requestNameLenLimit)
                 return False, msg
 
+        ## Check that --dryrun is not used with Automatic splitting
+        if getattr(self.configuration.Data, 'splitting', 'Automatic') == 'Automatic' and self.options.dryrun: 
+            msg = "The 'dryrun' option is not needed with the 'Automatic' splitting mode (default)."
```
belforte: not needed --> not compatible

PR #4772, opened 2018-06-20
Avoid usage of 'maxJobRuntimeMin' with Automatic splitting

### LINKED ISSUE #4771: beware of config.JobType.maxJobRuntimeMin in automatic splitting
I found that if the crabconfiguration indicates a value for config.JobType.maxJobRuntimeMin,
that value is then used in the processing DAG submission, which can lead to very obscure failure.
I do not suggest to fix this this in the automatic splitting code now, but we should
1. add to documenation
1. raise an error at crab submit time
which surely is easier and less risky than touching the automatic splitting code

lecriste:
Resolves #4771


belforte:
Looks good. Go ahead.

PR #4775, opened 2018-07-24
Reporting tw_name in status command.

todor-ivanov:


src/python/CRABClient/Commands/status.py
```diff
@@ -326,6 +327,9 @@ def printTaskInfo(self, crabDBInfo, username):
         if schedd:
             msg = "Grid scheduler:\t\t\t%s" % schedd
             self.logger.info(msg)
+        if twname:
+            msg = "Task Worker:\t\t\t%s" % twname
+            self.logger.info(msg)
```
lecriste: I prefer to not print this information in the crab status output and keep it only in the log, i.e. replace `self.logger.info` with `self.logger.debug`, in order to prevent user confusion with meaningless lines.
Same for the schedd name [three lines above](https://github.com/dmwm/CRABClient/pull/4775/files#diff-ff67af59351a4436cecb26ca2051f7b8R329).

todor-ivanov:
This is not for the users, but to facilitate the work of the operators. Of course hiding both the 'schedd' and the 'tw' from the users will make the 'status' command look prettier but sometimes digging into the the log files by the operators may cost more (it could even result in missing a problem just because it is blurred by the too many things in the logs).

belforte:
I agree with Todor, please find a compact way to list TW name in same line which we already have for schedd name sent from Stefano's mobileOn Jul 24, 2018 17:39, todor-ivanov <notifications@github.com> wrote:This is not for the users, but to facilitate the work of the operators. Of course hiding both the 'schedd' and the 'tw' from the users will make the 'status' command look prettier but sometimes digging into the the log files by the operators may cost more (it could even result in missing a problem just because it is blurred by the too many things in the logs).

todor-ivanov:
I tried to squeeze the two messages in a single line, but the infos about the schedd and the TW are arriving at different moments and are subject to different checks in the code. If I try to make it on a single line it will look differently every time a 'crab status' is executed and will make the code more ugly. I prefer it the way it is now (simple and readable). I will merge it like that and will create a tag for preprod  to be able to use the client while debugging the two TWs  in prod.

belforte:
@todor-ivanov you may find it easier to use your own client from git with all committs you like than going through a pre-prod tag, for sake of checking how things work now on production.
That said, sure, go ahead with what works for yoiu, I am opening an issue to clean up later.

PR #4780, opened 2018-09-06
Add CRIC support for DN to username resolution

belforte:


belforte:
@lecriste sorry Leonardo for having merged before you could review. I baiscally miss-clicked. Please review, test it and let's talk about it. We can always rolll this back. We surely want testing by various people before changing the client in CVMFS.

lecriste:
I reviewed the changes and they look good to me.
As a test I will run the following commands, is it enough?

- crab checkusername
- crab checkwrite
- crab uploadlog

belforte:
thanks Leonardo. Yes that should be enough, but I encourage all of us to use this client in all daily work. Maybe you can create a 1810.pre1 tag ?

Issue #4784, opened 2018-10-30
When AutomaticSplitting fails to create DAG, point to proper log


belforte:
see https://github.com/dmwm/CRABServer/issues/5751

try to detect tasks which failed whlie creating processing DAG, which are in the otherwise confusing status of scheduler FAILED but only listed jobs (probe ones) are all successful;
and point users (and operators!) to the proper log for details
```
web_dir:/AutomaticSplitting_Log0.txt
```

PR #4789, opened 2018-12-26
Improve crab status output for Automatic splitted tasks

lecriste:


src/python/CRABClient/Commands/status.py
```diff
@@ -569,14 +569,18 @@ def terminate(states, jobStatus, target='no output'):
...
                     self.logger.info("\nThe jobs splitting of this task is 'Automatic', please refer to this FAQ for a description of the jobs status summary:\n%s", automaticSplittFAQ)
                 total = sum(currStates[st] for st in currStates)
                 state_list = sorted(currStates)
                 self.logger.info("\n{0:32}{1} \t\t {2}".format(jobtype + ' status:', self._printState(state_list[0], 13), self._percentageString(state_list[0], currStates[state_list[0]], total)))
                 for jobStatus in state_list[1:]:
                     self.logger.info("\t\t\t\t{0} {1}".format(self._printState(jobStatus, 13), self._percentageString(jobStatus, currStates[jobStatus], total)))
+                if jobtype == 'Probe jobs':
+                    scheddName = self.schedd.replace("crab3@","")
+                    scheddNum = self.schedd.replace("crab3@vocms","").replace(".cern.ch","")
+                    self.logger.info("Probe stage log:\t\t%s", self.webdir.replace("http://","https://"+self.serverurl).replace(scheddName+"/mon","/scheddmon/"+scheddNum)+"/AutomaticSplitting_Log0.txt")
```
belforte: should be able to proxiedwebdir here, insread of this complicated URL editing

belforte:
can you give an examle of crab status output before/after this change ?

src/python/CRABClient/Commands/status.py
lecriste: OK thanks!

lecriste:
@belforte why would data discovery fail after probe step and not before?
The main reason for introducing the probe stage log is https://github.com/dmwm/CRABClient/issues/4784

lecriste:
I did not find an easy way to acess the lumis information.
I decided to add [this](https://github.com/dmwm/CRABClient/pull/4789/commits/eb08573ed3149a145ab41959f7cdf72a0d1b2bd6#diff-ff67af59351a4436cecb26ca2051f7b8R560) to differentiate main processing jobs from others.

belforte:
yes, the failure would be during splitting, not data discovery, sorry.
About lumis.. I know that it is not immediately available, OTOH the info is there since it is what crab report gets. Let's keep in mind as a possible improvement.

PR #4799, opened 2019-03-05
No max job runtime min with automatic splitt

lecriste:
@belforte shall I fetch this value:
https://github.com/dmwm/CRABServer/blob/631c55d6cc9f5aae1c73085b1e38051d00a4494c/src/python/TaskWorker/Actions/Splitter.py#L79
instead of redefining it here:
https://github.com/dmwm/CRABClient/compare/master...lecriste:no_maxJobRuntimeMin_with_AutomaticSplitt?expand=1#diff-71ffbfee13de0d1a2aef56c45be36b39R197
?

### DIFF
--- a/src/python/CRABClient/Commands/submit.py
+++ b/src/python/CRABClient/Commands/submit.py
@@ -217,7 +217,7 @@ class submit(SubCommand):
                 msg = "Invalid CRAB configuration: Parameter Data.unitsPerJob must be > 0, not %s." % (self.configuration.Data.unitsPerJob)
                 return False, msg
             if autoSplitt and (self.configuration.Data.unitsPerJob > autoSplittUnitsMax  or  self.configuration.Data.unitsPerJob < autoSplittUnitsMin):
-                msg = "Invalid CRAB configuration: In case of Automatic splitting, the Data.unitsPerJob parameter must be in the [%d, %d] minutes range. You askd for %d minutes." % (autoSplittUnitsMin, autoSplittUnitsMax, self.configuration.Data.unitsPerJob)
+                msg = "Invalid CRAB configuration: In case of Automatic splitting, the Data.unitsPerJob parameter must be in the [%d, %d] minutes range. You asked for %d minutes." % (autoSplittUnitsMin, autoSplittUnitsMax, self.configuration.Data.unitsPerJob)
                 return False, msg
         elif not autoSplitt:
             # The default value is only valid for automatic splitting!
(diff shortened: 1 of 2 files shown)


src/python/CRABClient/Commands/submit.py
```diff
@@ -192,13 +192,17 @@ def validateConfig(self):
                 msg = "Invalid CRAB configuration: Parameter General.requestName should not be longer than %d characters." % (requestNameLenLimit)
                 return False, msg
 
+        splitting = getattr(self.configuration.Data, 'splitting', 'Automatic')
+        autoSplitt = True if splitting == 'Automatic' else False
+        autoSplittUnitsMin = parametersMapping['on-server']['minAutomaticRuntimeMins']['default']
+        autoSplittUnitsMax = 2700 # 45 hours
```
lecriste: This parameter is not defined on the server so I did not add it to the `parametersMapping` dictionary, as I did for the one in the line above.
@belforte, do you like it defined here?

belforte:
@lecriste what does it mean "align with legacy convention" ?  Let's use comments that will make sense when read 3 years from now by someone else than you.

lecriste:
> @lecriste what does it mean "align with legacy convention" ?

I mean moving the parameter definition in the `parametersMapping` dictionary, since it is used also on the server. The commit code says it all:
https://github.com/dmwm/CRABClient/pull/4799/commits/5abdf1188fef44b083502ab4af790f4bd3055964

> Let's use comments that will make sense when read 3 years from now by someone else than you.

Anyway we always "Squash and merge" so this single commit message will not go into the master branch history.

belforte:
OK. I will wait for the PR to review comments !

src/python/CRABClient/Commands/submit.py
```diff
@@ -212,9 +216,12 @@ def validateConfig(self):
             if not int(self.configuration.Data.unitsPerJob) > 0:
                 msg = "Invalid CRAB configuration: Parameter Data.unitsPerJob must be > 0, not %s." % (self.configuration.Data.unitsPerJob)
                 return False, msg
-        elif getattr(self.configuration.Data, 'splitting', 'Automatic') != 'Automatic':
+            if autoSplitt and (self.configuration.Data.unitsPerJob > autoSplittUnitsMax  or  self.configuration.Data.unitsPerJob < autoSplittUnitsMin):
+                msg = "Invalid CRAB configuration: In case of Automatic splitting, the Data.unitsPerJob parameter must be in the [%d, %d] minutes range. You askd for %d minutes." % (autoSplittUnitsMin, autoSplittUnitsMax, self.configuration.Data.unitsPerJob)
```
belforte: You askd --> You asked

belforte:
OK for me. You can merge, possibly with the typo fix I indicated in previous comment.

PR #4808, opened 2019-04-18
Add Grafana dashboard link

lecriste:
@belforte, shall we print [this](https://github.com/dmwm/CRABClient/compare/master...lecriste:improve-crab_status?expand=1#diff-ff67af59351a4436cecb26ca2051f7b8R357) in a different color till the colon?

### DIFF
--- a/src/python/CRABClient/Commands/status.py
+++ b/src/python/CRABClient/Commands/status.py
@@ -480,7 +480,7 @@ class status(SubCommand):
             for param, values in usage.items():
                 if values[0] and values[1] > values[3]:
                     if values[0] < values[2]*values[1]:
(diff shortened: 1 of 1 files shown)


belforte:
I need to see what crab status output looks like. Will reply once I have done it.

belforte:
Keep it like this, better not have too many colors.
Screenshot from 2019-04-18 10-31-30

We will also need a message to the users in HN.

Rather, once we believe that new dashbaord is good to go, we invert things, list the new link first and put the old one as backup. Hopefully will do this in one month or so. Otherwiswe the July 1st that Valentin and Federica set for the termination of old dashbaord needs to be changed.

Issue #4809, opened 2019-04-24
Submission failing because of missing PSetDump.py


lecriste:
As discussed here:
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/4782/1/1/1/1/1.html

`crab submit` fails because can not find `PSetDump.py`.

Relevant log lines:

DEBUG 2019-04-24 09:36:04,367: 	 Writing CMSSW configuration to /afs/cern.ch/work/c/cerminar/egml1/dev/cc7/ntuple_prod/CMSSW_10_5_0/src/submission/ntuples/v00/SinglePhoton_FlatPt-8to150_PU0/crab_SinglePhoton_FlatPt-8to150_PU0/inputs/PSet.py
DEBUG 2019-04-24 09:36:04,985: 	 Cannot dump CMSSW configuration file. This prevents sandbox recycling but it is not a fatal error.
ERROR 2019-04-24 09:36:04,985: 	 Sequence has a label, but using its label to get an attribute from the process yields a different object or None
label = hgcalBackEndLayer2
DEBUG 2019-04-24 09:36:04,986: 	 The following EDM output files will be collected: []
...
DEBUG 2019-04-24 09:36:13,422: 	 Adding data directory /afs/cern.ch/work/c/cerminar/egml1/dev/cc7/ntuple_prod/CMSSW_10_5_0/src/L1Trigger/VertexFinder/interface to tarball
ERROR 2019-04-24 09:36:13,787: 	 Unhandled Exception!
ERROR 2019-04-24 09:36:13,787: 	 [Errno 2] No such file or directory: '/afs/cern.ch/work/c/cerminar/egml1/dev/cc7/ntuple_prod/CMSSW_10_5_0/src/submission/ntuples/v00/SinglePhoton_FlatPt-8to150_PU0/crab_SinglePhoton_FlatPt-8to150_PU0/inputs/PSetDump.py'
Traceback (most recent call last):
  File "/cvmfs/cms.cern.ch/crab3/slc6_amd64_gcc493/cms/crabclient/3.3.1903.patch1/bin/crab", line 160, in <module>
    client()
  File "/cvmfs/cms.cern.ch/crab3/slc6_amd64_gcc493/cms/crabclient/3.3.1903.patch1/bin/crab", line 148, in __call__
    self.cmd()
[...]

PR #4810, opened 2019-04-28
Resolve part of #4809

lecriste:
Fail submission when the dump of CMSSW configuration file cannot be done.

### DIFF
--- a/src/python/CRABClient/Commands/status.py
+++ b/src/python/CRABClient/Commands/status.py
@@ -480,7 +480,7 @@ class status(SubCommand):
             for param, values in usage.items():
                 if values[0] and values[1] > values[3]:
                     if values[0] < values[2]*values[1]:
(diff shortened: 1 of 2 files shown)


belforte:
I do not fully understand the `filterName` and PSET thing, but given that only Matthias had a problem there, and he left >1 year ago. I'd rather stay on the safe side and do not thouch this part.

belforte:
Even if CRAB code itself doe not look for  filterName, it is possible that this is required in order to make sure that output from different modules is written in the FJR with enough information for CRAB to be able to publish in different datasets as it is supposed to be able to.
A different story is as Matthias wrote "why is this required if there is no publication".
But lacking a full understanding of how this is used and why it was put there in the first place, and given that multiple output streams are something that user seldom use and we never test, I say that we do not touch this part of the code.
Pleae make a PR with only has the change for makiing the Pset Dump failure fatal.

PR #4821, opened 2019-08-26
Fix 4820

belforte:


src/python/CRABClient/ClientUtilities.py
```diff
@@ -190,9 +191,13 @@ def getColumn(dictresult, columnName):
 def getUrl(instance='prod', resource='workflow'):
     """
     Retrieve the url depending on the resource we are accessing and the instance.
+    As of August 2019 the Kubernetes server instance (k8s) points to preprod database instance
     """
     if instance in SERVICE_INSTANCES.keys():
-        return BASEURL + instance + '/' + resource
+        if instance == 'k8s':
+          return BASEURL + 'preprod' + '/' + resource
```
lecriste: Only two indent spaces? should be four

src/python/CRABClient/ClientUtilities.py
```diff
@@ -190,9 +191,13 @@ def getColumn(dictresult, columnName):
...
+        else:
+          return BASEURL + instance + '/' + resource
```
lecriste: Same comment as above

belforte:
@lecriste I resolved the conflict and did the modification which you requested earlier. Please check and if good merge. Thanks

Issue #4581, opened 2015-12-07
make sure input pset is not executed twice when building hash


belforte:
see
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/1191.html

belforte:
may be the same as #4493 which is very old and I do not know why was never fixed, maybe fell off radar

belforte:
I believe that it will be solved once the plan outlined here is implemented
https://github.com/dmwm/CRABClient/pull/4824#issuecomment-528646511

Issue #4493, opened 2015-05-21
avoid multiple CMSSW config imports when using CRAB CLIENT API


belforte:
see
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/461/1.html
and
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/622/1/1/2/1/1.html

AndresTanasijczuk:
and https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/665.html

belforte:
this was already in https://github.com/dmwm/CRABClient/issues/4248

belforte:
I believe that it will be solved once the plan outlined here is implemented
https://github.com/dmwm/CRABClient/pull/4824#issuecomment-528646511

PR #4824, opened 2019-09-02
Fixed caching of CMSSW Configuration

cbbrainerd:
`configurationCache` does not work as intended because `os.getcwd()` is appended to `sys.path` every time a configuration file is loaded. Since `sys.path` is used as part of the key, `cacheLine`, used to fetch a cached configuration, the configuration cache is never used.

Since `sys.path` is not changed anywhere else except in the (undocumented) external plugin feature there is no reason to use it as a key to the `configurationCache`, so it can just be removed from `cacheLine` and allow the `configurationCache` to be used as intended.

(Even if `sys.path` is changed reloading the file isn't the right thing to do anyway- reloading the same CMSSW configuration file multiple times can cause crashes and difficult-to-debug incorrect behavior)

### DIFF
--- a/src/python/CRABClient/JobType/CMSSWConfig.py
+++ b/src/python/CRABClient/JobType/CMSSWConfig.py
@@ -50,10 +50,14 @@ class CMSSWConfig(object):
             configFile, pathname, description = imp.find_module(cfgBaseName, [cfgDirName])
             cacheLine = (pathname, tuple(sys.argv))
             if cacheLine in configurationCache:
-                self.fullConfig = configurationCache[cacheLine]
+                if sys.path != configurationCache[cacheLine]['path']:
+                    self.logger.warning('Warning: sys.path has changed since CMSSW configuration file was loaded.')
+                self.fullConfig = configurationCache[cacheLine]['config']
                 configFile.close()
             elif not bootstrapDone():
-                sys.path.append(os.getcwd())
+                cwd=os.getcwd()
+                if cwd not in sys.path:
+                    sys.path.append(cwd)
                 try:
                     oldstdout = sys.stdout
                     sys.stdout = open(logger.logfile, 'a')
@@ -62,7 +66,7 @@ class CMSSWConfig(object):
                     sys.stdout.close()
                     sys.stdout = oldstdout
                     configFile.close()
-                configurationCache[cacheLine] = self.fullConfig
+                configurationCache[cacheLine] = { 'config' : self.fullConfig , 'path' : sys.path }
             self.logger.info("Finished importing CMSSW configuration %s" % (userConfig))
             sys.argv = originalArgv
 
--- a/src/python/CRABClient/JobType/CMSSWConfig.py
+++ b/src/python/CRABClient/JobType/CMSSWConfig.py
@@ -48,7 +48,7 @@ class CMSSWConfig(object):
                 msg = "Additional parameters for the CMSSW configuration are: %s" % (pyCfgParams)
                 self.logger.debug(msg)
             configFile, pathname, description = imp.find_module(cfgBaseName, [cfgDirName])
-            cacheLine = (tuple(sys.path), tuple(pathname), tuple(sys.argv))
+            cacheLine = (pathname, tuple(sys.argv))
             if cacheLine in configurationCache:
                 self.fullConfig = configurationCache[cacheLine]
                 configFile.close()


belforte:
thanks for this patch. I am not familiar with this part of the code and am a bit puzzled since the offending line had been put in there originally exactly to solve the problem of multiple reloading https://github.com/dmwm/CRABClient/commit/0795a60ab68db6031c7f902b3da487d504c1f115
I am not saying that current code is correct, only that I can't judge on this change by simply reading it. @cbbrainerd can you point to one example of a problem which this is fixing ? Does it maybeaddress #4493 or #4581 ?

cbbrainerd:
@belforte I believe it should fix #4581. At least it fixes a similar problem I was having with multiple loading.

The `configurationCache` code is meant to avoid the multiple loading, but because the path changes between the original invocation and the next the tuple used as a key to the dictionary is unique every time and so the cache is never actually used.

belforte:
thanks @cbbrainerd . Please allow me a bit to get familiar with the code here in order to review and test.

cbbrainerd:
@belforte Of course. Take your time.

Looking at the history, it appears the problem was indeed fixed by [0795a60](https://github.com/dmwm/CRABClient/commit/0795a60ab68db6031c7f902b3da487d504c1f115) but was later broken by [01c409c1](https://github.com/dmwm/CRABClient/commit/01c409c14968e7a1d66a823e582720760616b0cd).

belforte:
@mmascher this is a bit difficult for me to really understand and validate. Can you have a quick look and possibly a word of wisdom ?

belforte:
@PerilousApricot Hi Andrew, I know many years have gone, but can you have a look at what you wrote then and say why you used sys.path as part of the configuration cache here ? After I read the code I tend to agree with @cbbrainerd on the fact that it appears not needed, and it is quite possible that in many cases things work (better) w/o it. But one always has to fear that the code was there to address some use case which I am not thinking of at the moment.

belforte:
@cbbrainerd I may surely still miss a lot of things here, but I would think that there must have been some good reason for this change `Add cwd to sys.path before loading the pset. ` in https://github.com/dmwm/CRABClient/commit/01c409c14968e7a1d66a823e582720760616b0cd , even if Andres did not write the rational down and he has now left CMS since two years so we can't ask. I could imagina that maybe there's some place where a thing like `./pset.py` is executed and then the `cwd` needs to be in the path.
Anyhow, I was wondering if we can get the same result that you intended now, with the "safer" change from
```
sys.path.append(os.getcwd())
```
to 
```
if not os.getcwd() in sys.path:
    sys.path.append(os.getcwd())
```

cbbrainerd:
@lecriste That should work fine, as long as that line is added before the cache line is computed. If it is done after, then the cache line will still change be changed between the first time the configuration file is loaded and the second. Though it would be more fragile, since anyone changing the path elsewhere in the code would also break the multiple loading again.

I'm sure there was a good reason for changing `sys.path` before loading the pset, but I don't see why `sys.path` would ever need to be part of `cacheLine`.

We could instead add some code that checks to see if the path has changed, and issues a warning if so, like I've done here: [8366a9dd](https://github.com/cbbrainerd/CRABClient/commit/8366a9ddf420a6e08d872c5028e27b9129000e88).

As for testing, here is a minimal submit script using the crabCommand API that submits the same job twice:

```
from CRABClient.UserUtilities import config
from CRABAPI.RawCommand import crabCommand

config=config()
config.Site.storageSite='T3_US_FNALLPC'
config.JobType.pluginName='PrivateMC'
config.JobType.psetName='PSet.py'
config.Data.splitting='EventBased'
config.Data.unitsPerJob=1
config.Data.totalUnits=1
config.Data.outputPrimaryDataset='None'
config.Data.publication=False

crabCommand('submit',config=config)
crabCommand('submit',config=config)
```

with a corresponding PSet.py that crashes if it is imported twice:

```
import FWCore.ParameterSet.Config as cms
process = cms.Process('test')
process.source = cms.Source("EmptySource")
process.maxEvents = cms.untracked.PSet(input = cms.untracked.int32(1))

from crash import f

f()
```

using a python script crash.py to cause the crash:

```
def f():
    f.x-=1
    return 1/f.x
f.x=2
```

I can attach directions to a real workflow that has the same problem instead of a toy one, if you like, but the instructions are much more involved in that case.

belforte:
Thanks for the explanation and the example !
I'll look into this. Maybe the fear is that the
user may change PYTHONPATH between multiple
job submissions, e.g. changing CMSSW versions.
As usual it may have been better to exit with
error in case of rare, poorly debugged and
marginally useful use cases, but at times
developers get greedy and like the idea to
support all possibilities...

belforte:
I presume that the cached config is the things that's pickedl, i.e. not the python config that the user writes, but the fully expanded one after all things referred in there are imported from scram directory, which may be also user-modified things beside more stable cff's in the release. In this way the very same user config file could lead to different cached configurations if sys.path is changed, hence the need to change key.
But this is all wild guessing so far. I was not there when this stuff was discussed originally, designed, coded... I still believe such intricacies should belong to the layer above (i.e. the user). It may be better to ask the user to prepare a pickled config and tell CRAB whether it can be reused, taking responsibility for this actions, instead of trying to guess what he wanted to do and try to do it right even if she/he tried to do it wrong !
But we are not in a situation to entertain design changes.

cbbrainerd:
@belforte At any rate, I think the most sensible thing to do if the user changes `sys.path` between submissions is to warn that it may cause problems. The most common usage for the API I've seen is to simply submit the same job on multiple datasets, and multiple loading is definitely a problem in that case.

If someone wants to do something unusual, they can always fork the process and forgo the cache entirely as recommended as a short-term fix for this problem.

belforte:
++1

belforte:
trying to finalize...
   1. CRAB command API will not support changing $PYTHONPATH (i.e. sys.path) in the same multi-command process as e.g. changing CMSSW release would cause

**because**

   2. CMSSW configuration files should never be imported twice in the same process, since this is not supported by CMSSW and we know of cases where it led to immediate crash or (worse) obscure runtime errors

Current code tried to deal with 1. by loading a new config. in a new cache instance, leading to 2.
The only possible solution IMHO is that there can be only one cached configuration.

Therefore we should raise a FATAL ERROR, not a Warning.

If you agree I will take care and make sure that the above is reported in comments in the code, so that who will read in another 5 years may have an easier time.

Finally, as cleanup goes, the stuff about bootstrapDone is likely not needed/used since years and only bring confusion, but I have not looked at it enough to be sure yet that we can simply remove it.

belforte:
I am not sure what we have to do about this though
https://twiki.cern.ch/twiki/bin/view/CMSPublic/CRAB3FAQ#Multiple_submission_fails_with_a
i.e. the case that sys.path does not change, but user wants to change the pset parameters between calls.

PerilousApricot:
Sorry I missed this -- I have a filter on GH emails from DMWM because of the bot that sends out waytoo many mails. let me read this thread

PerilousApricot:
I don't understand the comment at the outset
```
configurationCache does not work as intended because os.getcwd() is appended to sys.path every time a configuration file is loaded. Since sys.path is used as part of the key, cacheLine, used to fetch a cached configuration, the configuration cache is never used.
```

 `os.getcwd` only changes if the working directory changes. How does that mean the configuration cache is never used?

belforte:
AFAICT it is added over and over every time this is called
becasue a new call to crab command API is done in a multicrab-like script

PerilousApricot:
In any case, you have to care about the current working directory, since sys.path can be populated with relative paths. See, for instance the zeroth element of sys.path:

```
# python
Python 2.7.5 (default, Oct 30 2018, 23:45:53) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-36)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> print sys.path
['', '/usr/lib64/python27.zip', '/usr/lib64/python2.7', '/usr/lib64/python2.7/plat-linux2', '/usr/lib64/python2.7/lib-tk', '/usr/lib64/python2.7/lib-old', '/usr/lib64/python2.7/lib-dynload', '/usr/lib64/python2.7/site-packages', '/usr/lib64/python2.7/site-packages/gtk-2.0', '/usr/lib/python2.7/site-packages']
```

PerilousApricot:
See the following reproducer:

```
[root@xrootd-vanderbilt ~]# mkdir test1 test2
[root@xrootd-vanderbilt ~]# echo "val = 'foo'" > test1/test.py
[root@xrootd-vanderbilt ~]# echo "val = 'bar'" > test2/test.py
[root@xrootd-vanderbilt ~]# python -c 'import os ; os.chdir("test1") ; import test ; print test.val'
foo
[root@xrootd-vanderbilt ~]# python -c 'import os ; os.chdir("test2") ; import test ; print test.val'
bar
```

The current cache wouldn't think the `test` module was the same, since the cwd changed, but with your patch, they could be incorrectly be considered the same. Correctness should trump speed in every case, IMO

belforte:
so we need to make sure that we check everything...
or in a different way: cache config the firstt time and always reuse.
the only think being how much effort to put to detect when user is overlooking this so we stop him from doing something that will not work

belforte:
let's state the requirements first. Will worry later about nitpicking the code.

PerilousApricot:
Well, currently, the code is just slower if they're doing something wrong. I think the safest option is to do something like

```python

def cached_import(filename)
    if filename in paths_already_loaded and not filename in caches:
        logger.warn("You're trying to import a previously-imported file %s but it's not in the cache. Please verify that you haven't changed the working directory or python's 'sys.path' variable")

    # keep importing
```

belforte:
any reason why it can't be a fatal error ? this can only happen in very odd edge cases

PerilousApricot:
Well, the code is still functional and provides the right results, it's just slower. IMO, the code should only bomb if there's no way to recover from the inputs

belforte:
basically the user can't change things in order to ask CRAB to load a new config. for next command, because loading anew config is exactly what we do not want to do. All that get's done in one "multicrab session" must use a single CMSSW config. We need to make this clear in the doc.

belforte:
current code as of latest commit is not really good. since by loading the config. twice it can lead to obscure runtime errors. I do not want to cache the thing again, if that's what you mean by slower

belforte:
I suspect there's disagreemen in  what "right result" means. It may mean "CRAB submits" or it may mean "CRAB submit a set of tasks which will behave dientically to multilple crab submit commands"

PerilousApricot:
Well, what I meant by "right result" is "gives the exact same code back" even if "CRAB submit a set of tasks which will behave dientically to multilple crab submit commands", but I forgot about the problem with CMSSW not being idempotent. In that case, it should just throw a fatal error and not a warning like your suggestion

belforte:
THanks @PerilousApricot . And thanks @cbbrainerd 
I feel like I know what to do here. Will try not to break what works now. And if OK with you all I will take care.

ONly if you care, you can read my plan below.

In the end there are two issues:
1. the old thing about no protection against CMSSW not being idempotent
2. `cacheline` variable is created before cwd is added to pythonpath, so at next iteration sys.path is not matching and config. is cached again . Leading to problem 1. even if user did not change anything.
the proposed solution is to add cwd to sys.path immediately and use the expanded $PYTHONPATH, the config name, its parameters (also the file hash ?) as 'config signature'. If in a later call we are asked to use a config with same signature, we reuse the pickle, otherwise throw fatal error. Even if  realistically the only time this could happen is in case user tries to change pset arguments. And simply code this in the most pedantic, stupid, but clear way.

PerilousApricot:
Wait, is cwd added to the pythonpath manually, or is it just the expansion
of "" in pythonpath?

>

belforte:
added manually. It was put there by Andres T. at some point.
Since he's gone since two years, I can't ask him why :-(

On 06/09/2019 10:54, Andrew Melo wrote:> Wait, is cwd added to the pythonpath manually, or is it just the expansion
> of "" in pythonpath?

belforte:
Easier for me is to merge this in and then do the minor modifications I'd like

PR #4827, opened 2019-09-10
Fix 4826

belforte:


belforte:
@lecriste please review and merge if OK

PR #4829, opened 2019-09-13
when submitting, print where to

belforte:
this will make it clear where task has been submitted to from crab submit stdout.
Mostly useful for our tests, but maybe also for users.


belforte:
example, when submitting to new Kubernetes server:

```
Finished importing CMSSW configuration demoanalyzer.py
Sending the request to the server at cmsweb-test.cern.ch
Success: Your task has been delivered to the k8s CRAB3 server.
Task name: 190913_163616:belforte_crab_20190913_183612
Project dir: ./crab_20190913_183612
```

PR #4854, opened 2019-11-07
Improve opesSSL check

lecriste:


src/python/CRABClient/Commands/SubCommand.py
```diff
@@ -240,8 +240,9 @@ def __init__(self, logger, cmdargs = None, disable_interspersed_args = False):
         opensslInfo = subprocess.check_output(["openssl", "version"]).strip('\n')
         self.logger.debug("OpenSSl version: %s", opensslInfo)
         opensslVersion = opensslInfo.split()[1]
-        if int(opensslVersion[0]) > 1 or (int(opensslVersion[0]) == 1 and int(opensslVersion[2]) > 0):
-            raise EnvironmentException("OpenSSl version > 1.0 are not supported: %s" % opensslVersion)
+        nDots = opensslVersion.count(".")
```
belforte: can you please write in English the logic that you want to implement here ?
Counting dots is hard to follow and likely fragile.
Why can't we use split('.')[1] to pick the middle figure in X.Y.Z ?
Maybe just a sanity check that we have such a string with 3 numbers and if not raise with "unknown version" ?

lecriste: > @lecriste can you please write in English the logic that you want to implement here ?
> Counting dots is hard to follow and likely fragile.
> Why can't we use split('.')[1] to pick the middle figure in X.Y.Z ?

We can but then you need a convoluted `if` condition like the [previous one](https://github.com/dmwm/CRABClient/pull/4854/files#diff-6af69d1178721ebd7a03d65cc9063323L243).

Instead with the [`rsplit`](https://github.com/dmwm/CRABClient/pull/4854/files#diff-6af69d1178721ebd7a03d65cc9063323R244) (which needs to know the number of dots in the string) we get the first part of the version string (X.Y) in one go and the if condition redues to `> 1`.
  
> Maybe just a sanity check that we have such a string with 3 numbers and if not raise with "unknown version" ?

I don't know whether any version has the same format (3 numbers).

belforte: just write first what you need and want to do, at the very least it will allow someone to understand the logic if a change is needed at ome point. As it is ,  I really do not know what exactly you want to do and whether it will wok. If there's no simpler ule, t may be better to explicitly check for the versions strings we know of  (3?) and raise on anything else. But if you want to protect against unknown format changes in the version string, counting dots will not do.

lecriste: > just write first what you need and want to do, at the very least it will allow someone to understand the logic if a change is needed at ome point. As it is , I really do not know what exactly you want to do and whether it will wok.

Ok I'll add a comment on the logic. 

> If there's no simpler ule, t may be better to explicitly check for the versions strings we know of (3?) and raise on anything else. But if you want to protect against unknown format changes in the version string, counting dots will not do.

Counting dots and right splitting the version string until the second last dot protects against any change may appear in the format after the second dot.

PR #4860, opened 2019-12-17
Resolve #4853

### LINKED ISSUE #4853: Limit requested memory to 10 GB
Client version of
https://github.com/dmwm/CRABServer/issues/5965

lecriste:


src/python/CRABClient/Commands/submit.py
```diff
@@ -247,6 +248,16 @@ def validateConfig(self):
             msg += " (i.e. will run %s job type)." % ("an analysis" if upper(crab_plugin_name) == 'ANALYSIS' else "a MC generation")
             self.logger.debug(msg)
 
+        ## Check that the requested memory does not exceed the allowed maximum.
+        nCores = getattr(self.configuration.JobType, 'numcores', 1)
```
lecriste: @belforte, is it necessary to specify a default also here (`1`) or is [this](https://github.com/dmwm/CRABClient/blob/a5775c3f10546d15071d5dc79086d84d5d0fe090/src/python/CRABClient/ClientMapping.py#L45) enough?

belforte: sorry I di not understand what (1) referes to.

lecriste: To the last argument of `getattr`, the only `1` in that line :)

belforte: let's keep the default for the getattr call.

belforte:
@lecriste I cannot merge this. It apparently brings in a lot of changes to src/python/CRABClient/JobType/CMSSWConfig.py which surely have nothing to do with #4853 
You should start clean from master, make a new branch and only make the few changes needed for that issue.

PR #4861, opened 2019-12-19
Add additional cmssw pythonpath if CRAB_CMSSW_SITE_PACKAGES is set

smuzaffar:
The new crab startup script https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_11_1_X/master/_crab-startup.file runs in clean env but it still needs to access CMSSW's `PYTHONPATH` for `crab submit` command. I propose that crab use `CRAB_CMSSW_SITE_PACKAGES` env variable to include extra cmssw python paths when it deals with CMSSW configuration. This change should not break crab's existing setup as no one should be setting `CRAB_CMSSW_SITE_PACKAGES` env variable. 
I was abe to submit a test job with this change plus https://github.com/cms-sw/cmsdist/pull/5442/files for new crab startup script

### DIFF
--- a/src/python/CRABClient/JobType/CMSSWConfig.py
+++ b/src/python/CRABClient/JobType/CMSSWConfig.py
@@ -48,6 +48,13 @@ class CMSSWConfig(object):
                 sys.argv.extend(pyCfgParams)
                 msg = "Additional parameters for the CMSSW configuration are: %s" % (pyCfgParams)
                 self.logger.debug(msg)
+
+            #Add cmssw python paths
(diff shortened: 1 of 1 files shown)


belforte:
looks good.

belforte:
@smuzaffar do you need a new tag in order to pull it into your build ?

smuzaffar:
yes please, if you can provide a tag then I can build and deploy it on cvmfs for more tests

belforte:
now in 3.3.1912.rc2

smuzaffar:
thanks @belforte

PR #4871, opened 2020-01-16
Revert part of #4865

lecriste:
Bug reported in
https://hypernews.cern.ch/HyperNews/CMS/get/computing-tools/5478.html


belforte:
looks good

Issue #4856, opened 2019-11-15
remove dependency from PandaServerInterface


belforte:
https://github.com/dmwm/CRABClient/blob/648a53075f3482f6151b14b011eaa4a6e9cf034e/src/python/CRABClient/JobType/Analysis.py#L16
and
https://github.com/dmwm/CRABClient/blob/648a53075f3482f6151b14b011eaa4a6e9cf034e/src/python/CRABClient/JobType/Analysis.py#L58
seems the only place where this PandaServer is used in all of CRAB, even if it is in CRABServer.

most likely there are other methods available/used in CRAB or WMCore to generate UUIDs, but otherwise better put this function in the common utilities file

https://github.com/dmwm/CRABServer/blob/fd28461c1cce6e26d8735e19969ec91f1b19257a/src/python/PandaServerInterface.py#L557-L570

belforte:
looks like everywhere else we use:
```
import uuid
        block_name = "%s#%s" % (dataset, str(uuid.uuid4()))
```
`uuid` module is only available from Ptyhon 2.5. So maybe original code tried to be compatible with python 2.4

belforte:
closed via https://github.com/dmwm/CRABClient/pull/4872

belforte:
In order to avoid adding PandaInterace to the buid, we need to change CRABServer
setup not to list this dependency anymore !

https://github.com/dmwm/CRABServer/blob/master/setup.py#L21-L25

Thanks @mmascher for having made me aware (again) that dependencies are listed in the setup script of the depending-from package, not in the setup of the package itself
https://github.com/dmwm/CRABClient/blob/master/setup.py

PR #4872, opened 2020-01-17
use uuid, remove dependency from PandaSI. Fix #4586

belforte:


belforte:
rats... I mistyped in the commit, this fixes #4856

PR #4900, opened 2020-03-31
Fixed validation of PSets in CMSSWConfig.py

cbbrainerd:
The current validation of PSets in CMSSWConfig.py uses ```if not getattr(...)``` in a few places to check if an attribute is defined. But ```getattr``` throws an AttributeError if the attribute is not defined, and so the helpful error message is not shown.

Adding a default of ```none``` keeps this from throwing when ```process``` and ```source``` are undefined, and does not change the behavior in other cases (e.g. when process or source are defined as ```None```).


belforte:
+1
thanks @cbbrainerd

PR #4924, opened 2020-06-25
make import of HTTPExceptions python3 compatible

belforte:


cbbrainerd:
@belforte This seems to break compatibility with older versions of python (The version 2.7.11 used in 8_0_X, at least), which gives the error 

```
  File "/cvmfs/cms.cern.ch/share/cms/crab-prod/v3.200719.00/bin/crab", line 25, in <module>
    from http.client import HTTPException
ImportError: No module named http.client
```

Is this intended? You could probably replace it with

```
try:
    from http.client import HTTPExceptions
except ImportError:
    from httplib import HTTPExceptions
```

without any real performance implications. Or add another python file that does this import and just import from there everywhere else it appears in the code.

belforte:
you are right. When I was suggested to do this change I was not told that it would only work for CMSSW >= 10.

Issue #4931, opened 2020-08-04
remove use of REST host DN in myproxy login hash


belforte:
given https://github.com/dmwm/CRABClient/issues/4927#issuecomment-660695253

may never need to get rid of the hash, but have to remove the REST name from it.
This will require changes also in TaskWorker and WMCore

- [x] change `proxy/getMyproxyUsernameForCRAB()` in WMCore to allow for missing `self.myproxyAccount`
- [x] get a new WMCore tag with it : this is `1.3.6.crab1` https://github.com/dmwm/WMCore/pull/9867 
- [x] change `CredntialInteractions/__init__()` in CRABClient not to add `myproxyAccount` (better to do it in the caller as well)
- [x] change  `CreateProxy/handleProxy)` in CRABClient not to pass `myproxyaccount` to `CredentialInteractions`
- [x] change TaskWorker to try both form of the hash https://github.com/dmwm/CRABServer/issues/6110
- [x] deploy new TW : TW version v3.200816 deployed in production on Aug 18
- [ ] build new CRABClient with those changes and new WMCore tag and deploy

belforte:
I tested all combinations of new TW with new/old client, existing credential, no credentia in myproxyl, clean /data/certs/creds in TW,... I hope it is OK now. Starting to roll out.
* TW pull request is ready https://github.com/dmwm/CRABServer/pull/6114
* need new WMCore
* once new TW is in production, can update CrabClient

belforte:
Now need to
- [ ] tag
- [ ] build
- [ ] test

belforte:
`crab-dev` now runs this and it is working.

PR #4933, opened 2020-08-06
No REST host name in DN hash for myproxy

belforte:


belforte:
Can only deploy once there is new WMCore and TaskWorker: https://github.com/dmwm/CRABClient/issues/4931#issuecomment-669974013

PR #4947, opened 2020-10-23
Use xroot for getcommand pfns

belforte:


belforte:
pylint complains are from pre-existing code, Let's not worry about them now.

belforte:
Pylint complains fixed. Ready for merging. @dciangot please comment whether you agree with this solution.

PR #4951, opened 2020-11-03
remove another call to OpenSSL, fix #4934

### LINKED ISSUE #4934: remove depenency from OpenSSL version
since we had to get back using DN hash in myproxy credential creation, we need a different
way to make Client forward compatible with newer OpenSSL since the solution in #4817 is not usable anymore and we had to re-introduce the check from  #4828

https://github.com/dmwm/CRABClient/blob/af5a38e069202412382aab4cc682c98f48585835/src/python/CRABClient/Commands/SubCommand.py#L241-L246

belforte:


belforte:
tested OK

PR #4962, opened 2020-11-16
Create tmp crab3checkwrite file only if lfn2pfn was successful (fix #4961)

### LINKED ISSUE #4961: make sure crab checkwrite cleans up
the are cases (e.g. a non existing site name) where the temp file which it creates locally
for the test is not cleaned up, that's ugly

@ddaina let me know if it is ok for you to do this, should be quite simple. the file is
https://github.com/dmwm/CRABClient/blob/master/src/python/CRABClient/Commands/checkwrite.py

you can wait when curent stuff you work on is done and take it as a test of my instructions in
https://twiki.cern.ch/twiki/bin/view/CMSPublic/CMSCrabClient#Develop_for_CRABClient

ddaina:
Create tmp crab3checkwrite file only if lfn2pfn was successful (fix #4961)

Tested using https://twiki.cern.ch/twiki/bin/view/CMSPublic/CMSCrabClient#Develop_for_CRABClient instructions.


belforte (approved):

belforte:
Perfect

PR #4964, opened 2020-12-01
try rucio.lfns2pfns with write first, than fallback to read

belforte:


belforte:
**to be tested**

belforte:
tested and OK

Issue #4971, opened 2021-02-11
replace CRABCache use with CERN S3


belforte:
most work will have to be done in the server.
CRABClient will have to replace the current uplaod to cache with two calls:
1. get a presigned URL from CRAB REST 
2.  upload the file that URL

all work of tracking object ids, checking used space, etc, will be done in CRABServer

Code to be changed is in here
https://github.com/dmwm/CRABClient/blob/59519827832e534bc6923af9972fcf44d5500e3a/src/python/CRABClient/ClientUtilities.py#L232-L259

belforte:
code that refers to crabcache in `crab purge` command can simply be commented-out/removed since we will deprecate that command (and remove the need for gsissh to schedd's !)

belforte:
Need to take care also of `dry run`:
https://github.com/dmwm/CRABClient/blob/59519827832e534bc6923af9972fcf44d5500e3a/src/python/CRABClient/Commands/submit.py#L376-L383

belforte:
@ddaina  I have started to work on this since I need at least to submit a task using S3 as cache to check what exactly goes in the DB and make progress on crabserver side. But I found that it is more work than initially thought since we need to pass a RESTServer object to the tarball handling code in 
https://github.com/dmwm/CRABClient/blob/59519827832e534bc6923af9972fcf44d5500e3a/src/python/CRABClient/JobType/UserTarball.py#L156-L159
The code I pointed to at the top of this issue is only about uploading log files :-(

So I decided to simplify a bit the way we handle the RESTServer communications now introducing a dictionary to track together the HTTPRequest object and the instance name:
```
self.REST = {'server': RESTServer, 'uriNoApi': uriNoApi}
```
this requires a change in almost all commands, as done earlier in
https://github.com/dmwm/CRABClient/commit/59519827832e534bc6923af9972fcf44d5500e3a

It is easier if I take care of this myself. I will leave the log upload part to you though.

belforte:
@ddaina I am fixing last pylint complains then will merge #4983 which , coupled with latest, unmerged yet, changes to the two files imported from CRABServer, manages to submit using S3 as  crabcache.
I will also merge and tag CRABServer as soon as I have done it on CRABClient.

Of course submitted task does not work, because TW does not know how to use S3 yet. I will work on that, in the meanwhile you can, in whichever order:
* look into uploadlog
[...]

PR #4983, opened 2021-04-08
Use s3 for cache. partial solution for #4971

belforte:


belforte:
requires new version of crabserver. wlll fail Jenkins at least until that.
But I need to take stock of changes so far and go with incremental fixes from now on

PR #4990, opened 2021-04-20
upload logfile to s3 (fix #4971)

### LINKED ISSUE #4971: replace CRABCache use with CERN S3

ddaina:
@belforte following changes will upload log file to s3 if `cacheSSL` for a given task in https://gitlab.cern.ch/crab3/CRAB3ServerConfig/-/blob/master/cmsweb-rest-config.json points to s3. otherwise logfile will be uploaded to old crabcache.

`uploadLog` to s3  only works with `crab uploadlog --dir` option. 
if `crab uploadlog --logpath` is used, log file is uploaded to old `crabcache` as in that case `cacheSSL` points to `https://cmsweb.cern.ch:8443/crabcache`. as already discussed, we should agree if `--logpath` option should be kept going forward

### DIFF
--- a/src/python/CRABClient/ClientUtilities.py
+++ b/src/python/CRABClient/ClientUtilities.py
@@ -21,7 +21,7 @@ from optparse import OptionValueError
 
 ## CRAB dependencies
 import CRABClient.Emulator
-from ServerUtilities import SERVICE_INSTANCES, uploadToS3
+from ServerUtilities import SERVICE_INSTANCES, uploadToS3, getDownloadUrlFromS3
 from CRABClient.ClientExceptions import ClientException, TaskNotFoundException, CachefileNotFoundException, ConfigurationException, ConfigException, UsernameException, ProxyException, RESTCommunicationException
 
 
@@ -222,8 +222,10 @@ def uploadlogfile(logger, proxyfilename, taskname=None, logfilename=None, logpat
         doupload = False
         
     if doupload:
-        logfileurl = False
-        logger.info("Uploading log file...")
+        if not username:
+            from CRABClient.UserUtilities import getUsername
+            username = getUsername(proxyFile=proxyfilename, logger=logger)
+
         # uploadLog is executed directly from crab main script, does not inherit from SubCommand
         # so it needs its own REST server instantiation
         restClass = CRABClient.Emulator.getEmulator('rest')
@@ -235,25 +237,24 @@ def uploadlogfile(logger, proxyfilename, taskname=None, logfilename=None, logpat
         # doesn't support unicode.
         cacheurl = cacheurl['cacheSSL'].encode('ascii')
 
+        logger.info("Uploading log file...")
         if 'S3' in cacheurl.upper():
-            #below line uploads log file to S3
-            uploadToS3(crabserver=crabserver, filepath=logpath, objecttype='clientlog', taskname=taskname, logger=logger)
+            objecttype = 'clientlog'
+            uploadToS3(crabserver=crabserver, filepath=logpath, objecttype=objecttype, taskname=taskname, logger=logger)
+            logfileurl = getDownloadUrlFromS3(crabserver=crabserver, filepath=logpath, objecttype=objecttype, taskname=taskname, username=username, logger=logger)
         else:
             cacheurldict = {'endpoint': cacheurl, "pycurl": True}
             ufc = UserFileCache(cacheurldict)
             logger.debug("cacheURL: %s\nLog file name: %s" % (cacheurl, logfilename))
             ufc.uploadLog(logpath, logfilename)
-            #logger.info("%sSuccess%s: Log file uploaded successfully." % (colors.GREEN, colors.NORMAL))
             logfileurl = cacheurl + '/logfile?name='+str(logfilename)
-            if not username:
-                from CRABClient.UserUtilities import getUsername
-                username = getUsername(proxyFile=proxyfilename, logger=logger)
             logfileurl += '&username='+str(username)
-            logger.info("Log file URL: %s" % (logfileurl))
+        logger.info("Log file URL: %s" % (logfileurl))
         logger.info("%sSuccess%s: Log file uploaded successfully." % (colors.GREEN, colors.NORMAL))
 
     else:
         logger.info('Failed to upload the log file')
+        logfileurl = False
 
     return  logfileurl
 
--- a/src/python/CRABClient/ClientUtilities.py
+++ b/src/python/CRABClient/ClientUtilities.py
@@ -21,7 +21,7 @@ from optparse import OptionValueError
 
 ## CRAB dependencies
 import CRABClient.Emulator
-from ServerUtilities import SERVICE_INSTANCES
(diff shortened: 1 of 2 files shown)


src/python/CRABClient/ClientUtilities.py
```diff
@@ -232,23 +234,26 @@ def uploadlogfile(logger, proxyfilename, logfilename=None, logpath=None, instanc
...
+            logger.info("Log file URL: %s" % (logfileurl))
         logger.info("%sSuccess%s: Log file uploaded successfully." % (colors.GREEN, colors.NORMAL))
-        logfileurl = cacheurl + '/logfile?name='+str(logfilename)
-        if not username:
-            from CRABClient.UserUtilities import getUsername
-            username = getUsername(proxyFile=proxyfilename, logger=logger)
-        logfileurl += '&username='+str(username)
-        logger.info("Log file URL: %s" % (logfileurl))
-        return  logfileurl
+
```
belforte, line 254: I am puzzled by the removal of this block of code. Was there some bug ?

belforte:
It looks to me that when using S3 the code will return `logfileurl=False`. Won't this confuse the caller ?

src/python/CRABClient/ClientUtilities.py
ddaina: I did not remove it, but moved it under `else` block.

belforte: oops

belforte:
Anyhow I pulled code from your branch, tested and it works.
So I simply do not know how `logfileurl` is really used in the code.

ddaina:
> It looks to me that when using S3 the code will return logfileurl=False. Won't this confuse the caller ?

that's true. I applied the logic which that was available before in case `uploadlog` command failed, then `logfileurl = False` is returned. so given that s3 does not return url, I set it to be `false`.

however, I noticed that I have overlooked and did not change this part which uses `uploadlogfile`:
https://github.com/dmwm/CRABClient/blob/a5c9b457d1981cba649f8b63f6a631ef1eb83297/bin/crab#L121-L132

belforte:
.hmmm... so this is what I got when I tested. Maybe simply the log file url is not printed anymore.
But I would expected this line to print something anyhow, or throw error
https://github.com/dmwm/CRABClient/blob/a5c9b457d1981cba649f8b63f6a631ef1eb83297/bin/crab#L126
will investigate
```
belforte@lxplus732/TJ> crab uploadlog 
Will upload file /afs/cern.ch/work/b/belforte/CRAB3/TC3/TJ/crab_20210420_230909/crab.log.
Fetching user enviroment to log file
Uploading log file...
Success: Log file uploaded successfully.
Log file is /afs/cern.ch/work/b/belforte/CRAB3/TC3/TJ/crab_20210420_230909/crab.log
belforte@lxplus732/TJ>
```

belforte:
ah.. got it. That line #126 and the code part which refers to `logfileurl` is only relevant for the case where the log was uploaded automatically.  Which I had not tested. Need to figure out how to make it happen in a simple way.
Anyhow we should decide if to change and do not print the URL, or if to get a PreSignedUrl via `getDonwloadUrlFromS3`

belforte:
all in all I like the latter option more:
* get a PreSignedUrl via `ServerUtilities/getDonwloadUrlFromS3` and report it

while of course this will be not critical as soon as we fix crabserver/ui to work with S3

belforte:
OK. I put a spurious `raise` in one crab command and indeed log now has
```
INFO 2021-04-21 07:26:45.074 UTC: 	 Uploading log file...
DEBUG 2021-04-21 07:26:45.074 UTC: 	 Failed to upload log file automatically
```

Indeed crab client tries hard to catch all possible exceptions and report sensible error messges, this part of the code is only activated when there are genuine bugs which trigger unhandled exceptions.

src/python/CRABClient/ClientUtilities.py
```diff
@@ -232,20 +236,22 @@ def uploadlogfile(logger, proxyfilename, logfilename=None, logpath=None, instanc
...
-        logger.info("%sSuccess%s: Log file uploaded successfully." % (colors.GREEN, colors.NORMAL))
-        logfileurl = cacheurl + '/logfile?name='+str(logfilename)
-        if not username:
-            from CRABClient.UserUtilities import getUsername
-            username = getUsername(proxyFile=proxyfilename, logger=logger)
-        logfileurl += '&username='+str(username)
+        if 'S3' in cacheurl.upper():
+            objecttype = 'clientlog'
+            uploadToS3(crabserver=crabserver, filepath=logpath, objecttype=objecttype, taskname=taskname, logger=logger)
+            logfileurl = getDownloadUrlFromS3(crabserver=crabserver, filepath=logpath, objecttype=objecttype, taskname=taskname, username=username, logger=logger)
```
belforte, line 244: @ddain you should not need to send username nor filepath here. If it did not work w/o either it means there's something to fix.
The crabserver REST will get the username from the CMSWEB FrontEnd.
So you should not need the call to getUserName which you inserted above.
Please try and let me know if you have problems.

ddaina: you are right, it works. thank you!

belforte (approved):

belforte:
LGTM now.
I tested it on lxplus and works finely:
```
Will upload file /afs/cern.ch/work/b/belforte/CRAB3/TC3/LOCAL/crab_20210414_225550/crab.log.
Fetching user enviroment to log file
Uploading log file...
Log file URL: http://s3.cern.ch/CRABCacheDev/belforte/210414_205557%3Abelforte_crab_20210414_225550/clientlog?AWSAccessKeyId=ab00fcc889f248c495d2875f15a3badb&Expires=1621676791&Signature=JjhnYCx9M4GySYWPvz6RC5USxsw%3D
Success: Log file uploaded successfully.
Log file is /afs/cern.ch/work/b/belforte/CRAB3/TC3/LOCAL/crab_20210414_225550/crab.log
belforte@lxplus790/LOCAL> 
```
I leave it to you to merge and tag. We also need a new crabserver tag so we can get a new client build in IB

belforte:
I even managed to shorten the original url [1] to http://cern.ch/go/WMx8

when this is all over and done, we can look into doing that in the code and only report the short one.
But will need to make sure that CERN Web Services people have not problem with the number of URLs which will be generated.

 [1] http://s3.cern.ch/CRABCacheDev/belforte/210414_205557%3Abelforte_crab_20210414_225550/clientlog?AWSAccessKeyId=ab00fcc889f248c495d2875f15a3badb&Expires=1621676791&Signature=JjhnYCx9M4GySYWPvz6RC5USxsw%3D

PR #4994, opened 2021-04-22
remove automatic log upload (fix #4993)

### LINKED ISSUE #4993: remove automatic log uploading
in light of replacing `UFC` with `S3`, we should remove automatic log upload in:
https://github.com/dmwm/CRABClient/blob/a5c9b457d1981cba649f8b63f6a631ef1eb83297/bin/crab#L121-L132
there is no way to access info if cache is `S3` or old `UFC` and  upload log file to `S3` from this code. exceptions should be catched while inside crab sub-command execution and crab uploadlog should be used then.

ddaina:


belforte (approved):

belforte:
LGTM  please test and merge. We can leave to another time to fix pylint complains in /bin/crab but if you feel like addressing at least some of them, it would be nice. at least W1201 should be easy,

PR #5008, opened 2021-05-28
protect against missing uberftp and fix #5007

### LINKED ISSUE #5007: crab checkwrite should not fail if uberftp is missing
some how in  https://github.com/dmwm/CRABClient/commit/77acae3f17e1792e85adb10ca4bfa14338a50c87 the check for CERN EOS username mapping via uberftp was unindented so it is called even if check is successful.
This is overkill which makes things uselessly verbose.
An in any case, users may be indicating T2_CH_CERN from any place where uberftp may actually be missing, not only lxplus.

Should only call `uberftp` when needed, check first if it exists, and if not tell user to log on lxplus and try there.

belforte:


belforte:
Jenkins functional test failed due to missing proxy. But I tested this on lxplus (has uberftp_ and crab-prod-tw01 (does not have uberftp)in both cases : test write to cern OK and test write to cern FAIL.

PR #5018, opened 2021-06-09
Py3 compat

belforte:


belforte:
will worry about pylint later, several complaints are about py3 constructed inside `if py3`.  I ran some tests locally in python2 and it seems to work. Merge and run client validation.

Issue #5014, opened 2021-06-07
uploadlog uses wrong REST instance


belforte:
maybe a problem introduced with the change for S3 ?
But with a task which is on the dev DB, crab uploadlog still tries to get server info from prod.

```
belforte@lxplus702/TC3> crab status
CRAB project directory:		/afs/cern.ch/work/b/belforte/CRAB3/TC3/crab_20210607_142031
Task name:			210607_122042:belforte_crab_20210607_142031
Grid scheduler - Task Worker:	crab3@vocms0107.cern.ch - crab-dev-tw01
Status on the CRAB server:	SUBMITTED
Task URL to use for HELP:	https://cmsweb-test2.cern.ch/crabserver/ui/task/210607_122042%3Abelforte_crab_20210607_142031
Dashboard monitoring URL:	https://monit-grafana.cern.ch/d/cmsTMDetail/cms-task-monitoring-task-view?orgId=11&var-user=belforte&var-task=210607_122042%3Abelforte_crab_20210607_142031&from=1623064842000&to=now
Status on the scheduler:	SUBMITTED

Jobs status:                    transferring 		100.0% (1/1)

No publication information (publication has been disabled in the CRAB configuration file)

Warning: the max jobs runtime is less than 30% of the task requested value (60 min), please consider to request a lower value for failed jobs (allowed through crab resubmit) and/or improve the jobs splitting (e.g. config.Data.splitting = 'Automatic') in a new task.

Warning: the average jobs CPU efficiency is less than 50%, please consider to improve the jobs splitting (e.g. config.Data.splitting = 'Automatic') in a new task

Summary of run jobs:
 * Memory: 6MB min, 6MB max, 7MB ave
 * Runtime: 0:04:05 min, 0:04:05 max, 0:04:05 ave
 * CPU eff: 20% min, 20% max, 20% ave
 * Waste: 0:06:09 (60% of total)

Log file is /afs/cern.ch/work/b/belforte/CRAB3/TC3/crab_20210607_142031/crab.log
belforte@lxplus702/TC3> crab uploadlog -d crab_20210607_142031
Will upload file /afs/cern.ch/work/b/belforte/CRAB3/TC3/crab_20210607_142031/crab.log.
Fetching user enviroment to log file
Fatal error trying to connect to https://cmsweb-test2.cern.ch:8443/crabserver/prod/info using {'subresource': 'backendurls'}
The server answered with an error.
Server answered with: No such instance
[...]

PR #5019, opened 2021-06-10
fix uploadlog for S3. Solve #5014

belforte:


belforte:
tested on lxplus. all OK.

PR #5029, opened 2021-06-14
use different fileopen mode for pickle for py2/3. Fix #5025

### LINKED ISSUE #5025: enforce uniform use of pickle
in python3 it makes a difference if file where we write pickle is text of binary
(opened with `rb/wb` or `r/w`)
need to be clear and consistent about this and encode/decode things that we pickle as needed.

For compatibility with pickle files written by current python2 code, should always use text mode.
I.e open in mode `r/w`

belforte:


belforte:
No. I made some branch mess and this reverts needed changes for Py3 compatibility. Will make a new PR

PR #5056, opened 2021-07-13
replace WMCore usage with curl (fix #4972)

### LINKED ISSUE #4972: replace use of CRABServer RESTInteraction class with plain `curl`
needed to remove dependency on WMCore

there are few calls done to CRABServer REST , usually one or two at most for each command execution,
there is no need for efficiency, speed, client reuse etc. so can replace RESTInteraction and WMCore's pycurl_manger with a simple class that prepares proper URL's (like RESTInteractions does) and used them by forkin a subprocess which executes `curl`.
Will also pave to way to the possibility of replacing `curl` with some static GOlang executable which solves all issues with system curl, openssl, SL6/7/8/... 

mostly and excise in python subprocess !

ddaina:
@belforte , this is first working version for discussion

however, my worry is: how to properly handle this part https://github.com/dmwm/CRABClient/blob/master/src/python/CRABClient/Commands/submit.py#L134-#L136 where given on current [implementation](https://github.com/dmwm/CRABClient/compare/master...ddaina:rmvWMCoreDependency?expand=1#diff-3a235bc3a0a3f4716cc9774eaba7e5e0986d4c09f915f9b342de13229f94fcefR145) if http code != 200, it will even not reach that code part.

Be aware that PR test will not test these changes because tests uses old RESTInteractions file for deploying CRABClient.

### DIFF
--- a/src/python/CRABClient/CrabRestInterface.py
+++ b/src/python/CRABClient/CrabRestInterface.py
@@ -31,10 +31,10 @@ EnvironmentException = Exception
 
 def retriableError(http_code, curlExitCode):
     """
-        checks if the error can be retried
+        checks if error is worth retrying
         :param http_code: int : HTTP code form the HTTP call if it was possible to do the call and obtain it
         :param curlExitCode: int : exit code of the curl command that was forked to execute the HTTP call
-        :return: True of False indicating if the error can be retried
+        :return: True of False indicating if the error is worth retrying
     """
     retry = False
 
--- a/src/python/CRABClient/CrabRestInterface.py
+++ b/src/python/CRABClient/CrabRestInterface.py
@@ -1,0 +1,220 @@
+"""
+Handles client interactions with remote REST interface
+"""
+
+import os
+import time
+import random
+import subprocess
+from ServerUtilities import encodeRequest
+import json
+
+try:
+    from urllib import quote as urllibQuote  # Python 2.X
+except ImportError:
+    from urllib.parse import quote as urllibQuote  # Python 3+
+
+import logging
+
+try:
+    from TaskWorker import __version__
+except:  # pylint: disable=bare-except
+    try:
+        from CRABClient import __version__
+    except:  # pylint: disable=bare-except
+        __version__ = '0.0.0'
+
+EnvironmentException = Exception
+
+class HTTPRequests(dict):
+    """
+    This code is a simplified version of WMCore.Services.Requests - we don't
+    need all the bells and whistles here since data is always sent via json we
+    also move the encoding of data out of the makeRequest.
+
+    HTTPRequests does no logging or exception handling, these are managed by the
+    Client class that instantiates it.
+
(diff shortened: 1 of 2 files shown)


belforte:
do you worry about raising Exception rather than RESTCommunicationException ? Or what else ?

belforte:
why we don't get curl errors anymore ?
https://github.com/dmwm/CRABServer/blob/109b0f98aa41acbce128c53561cea53685da2e85/src/python/RESTInteractions.py#L44-L47
I would have kept the `retriableError` function, makes the code more readable and cleanly separates the
what-to-do part from the implementation details.
Th exitcode (`process.returncode`) is never checked and never used. I think we can not assume that it will always return 0.

belforte (changes requested):
see comments

src/python/CRABClient/CrabRestInterface.py
```diff
@@ -0,0 +1,244 @@
...
+        command += ' --key "%s"' % self['key']
+        command += ' --capath "%s"' % caCertPath
+        command += ' "%s"' % url
+
+        # retries this up at least 3 times, or up to self['retry'] times for range of exit codes
+        # retries are counted AFTER 1st try, so call is made up to nRetries+1 times !
+        nRetries = max(2, self['retry'])
+        for i in range(nRetries +1):
+            stdout, stderr, exitcode = ' ,99999', None, 99999
+	    #variable is not used by the client code, however it is kept to main old behaviour
```
belforte, line 160: main --> mimic

src/python/CRABClient/CrabRestInterface.py
```diff
@@ -0,0 +1,244 @@
...
+
+    if http_code in [429, 500, 502, 503] or exitcode in [28, 35]:
+        retry=True
+
+    return retry
+
+
+class HTTPRequests(dict):
+    """
+    This code is a simplified version of WMCore.Services.Requests - we don't
```
belforte, line 53: this initial description should be updated

src/python/CRABClient/CrabRestInterface.py
```diff
@@ -0,0 +1,244 @@
...
+except:  # pylint: disable=bare-except
+    try:
+        from CRABClient import __version__
+    except:  # pylint: disable=bare-except
+        __version__ = '0.0.0'
+
+EnvironmentException = Exception
+
+def retriableError(http_code, exitcode):
+    """ Return True if the error can be retried
```
belforte, line 33: indicate in the comments what the argumets are: http_code is the code form the http call if it was possible to do the call and get it, otherwise exitcode is the exit code of the curl command that was forked to execute the call. All in all this funciont is now specific of using curl to make the call. Maybe rename exitcode to curlExitCode ?

belforte (changes requested):
code looks good, but I suggest to improve a bit the comments to reflect the changes from old CRABInterface. See comments.

ddaina:
thanks @belforte. I've updated comments

belforte (approved):

Issue #5067, opened 2021-08-18
fix curl error reporting


ddaina:
after following changes #5056 were merged, user does not get error details reported in sufficient detail to understand what actually has happened. For example, trying to kill task that is in status `SUBMITFAILED` results in the following output using `crab-dev` client (the one which uses `curl` instead of `pycurl_manager`:

```
9:42:01 Killing task: /tmp/crabStatusTracking/crab_20210804_094133
09:42:04 ERROR: Exception: Fatal error trying to connect to https://cmsweb-testbed.cern.ch:8443/crabserver/preprod/workflow using workflow=210804_074136%3Acmsbot_crab_20210804_094133. Error details: 
09:42:04 curl: (22) The requested URL returned error: 500 Internal Server Error
09:42:04  
09:42:04 
09:42:04 	Please email hn-cms-computing-tools@cern.ch for support with the crab.log file or crab.log URL.
09:42:04 	Client Version: v3.210714
09:42:04 	Please use 'crab uploadlog' to upload the log file /tmp/crabStatusTracking/crab_20210804_094133/crab.log to the CRAB cache.
```

while current production client same operation on `SUBMITFAILED` tasks reports in the following way:
```
09:34:24 Killing task: /tmp/crabStatusTracking/crab_20210804_092848
09:35:26 Fatal error trying to connect to https://cmsweb-testbed.cern.ch:8443/crabserver/preprod/workflow using workflow=210804_072856%3Acmsbot_crab_20210804_092848. Error details: {'Content-Length': '745', 'X-Error-Http': '500', 'X-Rest-Time': '53973.913 us', 'Server': 'CherryPy/17.4.0', 'Connection': 'close', 'X-Error-Detail': 'Execution error', 'CMS-Server-Time': 'D=76619 t=1628062526710493', 'X-Rest-Status': '403', 'Date': 'Wed, 04 Aug 2021 07:35:26 GMT', 'Content-Type': 'text/html;charset=utf-8', 'X-Error-Info': 'You cannot kill a task if it is in the SUBMITFAILED status', 'X-Error-Id': '2512368d5214188cad28fe11ef04df27'}
09:35:26 The server answered with an error.
09:35:26 Server answered with: Execution error
09:35:26 Reason is: You cannot kill a task if it is in the SUBMITFAILED status
09:35:26 Error Id: 2512368d5214188cad28fe11ef04df27
09:35:26 Log file is /tmp/crabStatusTracking/crab_20210804_092848/crab.log
[...]

PR #5069, opened 2021-08-24
fix curl error reporting (#5067)

ddaina:


belforte:
I find the name `RestDataProcessingException` a bit too long and misleading.
What about 'RESTInterfaceException' ?

belforte (approved):

belforte:
New code looks good to me, but since this is about dealing with errors, and errors are difficult to (re)produce, we should plan for some extended test time before replacing current server access via crabserver/RESTInteraction

ddaina:
thanks @belforte for review

PR #5082, opened 2021-09-09
unlink file with error info (fix #5079)

### LINKED ISSUE #5079: getlog in py3 version retrieves one extra file
for a job which completed OK at first iteration,
crab getlog in v3.210903 retrieves not only the correct job_out.1.0.txt
 but also tried to fetch a non existing job_out.1.1.txt  which has content

```
belforte@lxplus7108/results> cat job_out.1.1.txt
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>404 Not Found</title>
</head><body>
<h1>Not Found</h1>
<p>The requested URL /mon/cms1627/210906_204833:belforte_crab_20210906_224825/job_out.1.1.txt was not found on this server.</p>
<hr>
<address>Apache/2.4.6 (CentOS) Server at vocms059.cern.ch Port 80</address>
</body></html>
belforte@lxplus7108/results> 
```

if  is possible that current code does not know how many times jobs ran and tries to retrieve iter 0, 1, 2.... until it gets error. In previous implementation error meant also no file. Now a file with the http returned message is print instead.

ddaina:
Once `curl` tries to get file which does not exists (http 404 error), details will still be written to `filename`. However, details are logged and file is removed.

alternative was to add `-f` flag to `curl` command and in this case file with error details would not be created at all, but as per `curl` documentation:

> In normal cases when an HTTP server fails to deliver a document, it returns an HTML document stating so (which often also describes why and more). This flag will prevent curl from outputting that and return error 22

this flag will prevent `curl` from outputting error, so we loose this information (while we want to keep it)

alternative to `-f` flag would be to use `--fail-with-body`, which:

> This flag will still allow curl to output and save that content but also to return error 22.  This is an alternative option to -f, --fail which makes curl fail for the same circumstances but without saving the content. 

however, this option is available in `curl` 7.76.0 version, while we are using older one.

changes tested on lxplus.

### DIFF
--- a/src/python/CRABClient/UserUtilities.py
+++ b/src/python/CRABClient/UserUtilities.py
@@ -120,11 +120,21 @@ def curlGetFileFromURL(url, filename = None, proxyfilename = None, logger=None):
     if logger:
         logger.debug("Will execute:\n%s", downloadCommand)
     stdout, stderr, rc = execute_command(downloadCommand, logger=logger)
+    errorDetails = ''
+
+    if rc != 0:
+        os.unlink(filename)
+        httpCode = 503
+    else:
+        httpCode = int(stdout)
+        if httpCode != 200:
+            with open(filename) as f:
+                errorDetails = f.read()
+            os.unlink(filename)
     if logger:
-        logger.debug('exitcode: %s\nstdout: %s\nstderr: %s', rc, stdout, stderr)
-    httpCode = int(stdout)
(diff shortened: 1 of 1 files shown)


belforte (approved):

belforte:
looks very good. thanks

belforte:
let's wait for Jenkins before merging, though

PR #5092, opened 2021-10-20
better message when *proxy* command fails. Fix #5091

### LINKED ISSUE #5091: more clearly report invalid proxies
In the new ProxyInteractions.py when things go wrong it simply prints to console
```
Problems getting voms proxy info.
```
with details being only in log.
But this is very likely to confuse users and end up in them sending thousand lines long logs to support, better to tell them more clearly where the problem is.
Older version used to force a new proxy creation in such case, but that's also not optimal since when it was failing it was usually due to something int he site setup which has nothing to do with crab and I do not want to read crab logs to find that they user environment is wrong to begin with.

belforte:


belforte:
thanks @ddaina , please run Jenkis from the HEAD again so we give it a good shake

Issue #4979, opened 2021-03-15
remove references to RESTInteractions and UserFileCache in Emulator.py


belforte:
once those are not used anymore in the code, there will be no need to define them in
https://github.com/dmwm/CRABClient/blob/master/src/python/CRABClient/Emulator.py

ddaina:
this is solved

PR #5103, opened 2021-11-03
remove reference to UserFileCache

ddaina:
temporary fix before https://github.com/dmwm/CRABServer/issues/6776


ddaina:
resolves #4979

belforte:
this is a good stepping stone, but please open another issue about changing all places where the  dictionary returned by `Emulator/getDefault` is used so that they `'ufc'` key can be remove instead of being assigned a dummy value. We can do that later after things have converged, of course.

belforte (approved):

ddaina:
thank you. sure, will open

Issue #4975, opened 2021-02-11
stop using WMCore Lexicon and Lumilist


belforte:
use of [Lexicon](https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/Lexicon.py) appears limited to a single regexp for run ranges.
- [x] make sure thenre's no other use
- [x] replace with a regexp in our code

more investigation  may be needed for
```
from WMCore.DataStructs.LumiList import LumiList
```
that is already py2/3 compatible, and likely to stay. But maybe there's something like [LumiList](https://github.com/dmwm/WMCore/blob/master/src/python/WMCore/DataStructs/LumiList.py) in CMSSW and since we run inside a CMSSW env. already, could switch to use [that](https://github.com/cms-sw/cmssw/blob/master/FWCore/PythonUtilities/python/LumiList.py) , but need to understand if I can really mix code in this way.
Here I'd rather not make a 3rd copy !

belforte:
@ddaina I am assigned to you as part of the plan we agreed on that you take ownership of making CRABClient py2/3 compatible.
I think the part about Lexicon is trivial and you will have no problem.
FOr the lumilist thing, let's talk about it when you will be ready to tackle it.
We can surely start here like for Configuration by picking the older version from WMCore which does not import `future`, builtins` and other too modern stuff.

mapellidario:
The main changes I added to WMCore.Configuration are

```python
from builtins import object, int, str as newstr, bytes as newbytes  # these come from future as well

from future.utils import listvalues
```

- `object` helps with py2/py3 compatibility  of class objects. Importing `object` from `future` can be skipped in our case
- the use of `listvalues` can be replaced by `list(value.values())` if we only use py3.
- importing `int`, `newstr` and `newbytes` it has been necessary to make other code using WMCore.Configuration in WMCore work. This module is used pretty extensively around WMCore and every little error here causes everything else to crumble, as this experiment suggest https://github.com/dmwm/WMCore/pull/10839#issuecomment-929958761 . It is definitely possible that CRABClient has different needs than WMCore and that we can simply use here a not-modernized version of Configuration.

Do not evaluate the py3 compatibility of WMCore.Configuration in absolute terms, it should be treated as a blueprint for what it serves. If WMCore uses extensively newstr and newbytes, then WMCore.Configuration should reflect it. If CRABclient has different needs, then it is reasonable to fork WMCore.Configuration and have our own (simplified?) version of it.

Hope this helps!

[...]

PR #5107, opened 2021-11-08
use lumilist from CMSSW

ddaina:
tested using CMSSW_7_6_7, CMSSW_10_2_6 CMSSW_12_0_1 releases


ddaina:
resolves #4975 part about lumilist usage

belforte (approved):
FIne

PR #5111, opened 2021-11-12
improve HTTP header parsing

ddaina:
brings two changes/improvements:
1. bin/crab.py: remove check if `exitcode==503` as this is repetitive and saves us from parsing response header which is already done in CrabRestInterface.py
2. CrabRestInterface.py: reuse example from WMCore pycurl_manager how to parse HTTP response header to get HTTP code. also, now regex used in this parsing is compatible both with the header return by curl and gocurl.

tested on CMSSW_7_6_7, CMSSW_10_2_6, CMSSW_12_0_1

@belforte these are last changes that i would like to merge before creating new tag and updating IB.

### DIFF
--- a/bin/crab.py
+++ b/bin/crab.py
@@ -149,11 +149,11 @@ if __name__ == "__main__":
         client()
         exitcode = 0 #no exceptions no errors
     except RESTInterfaceException as err:
+        exitcode=err.exitcode
         client.logger.info("The server answered with an error.")
         client.logger.debug("")
         err = str(err)
-        exitcode = int(re.findall(r'(?<=\<\sHTTP/1.1\s)[^\n\s]*', err)[-1])
-        if exitcode==503 and ("CMSWEB Error: Service unavailable") in err:
+        if ("CMSWEB Error: Service unavailable") in err:
             client.logger.info(schedInterv)
         if 'X-Error-Detail' in err:
             errorDetail = re.search(r'(?<=X-Error-Detail:\s)[^\n]*', err).group(0)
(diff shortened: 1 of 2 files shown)


bin/crab.py
```diff
@@ -149,11 +149,11 @@ def log_exception(exc_type, exc_value, tback):
         client()
         exitcode = 0 #no exceptions no errors
     except RESTInterfaceException as err:
+        exitcode=err.exitcode
```
belforte, line 152: I though that this syntad is not python3 compatible, i.e. exceptions do not have members anymore. But maybe I am confused, since I guess you tested. But for my educatin, @mapellidario can you chime in ?

belforte (approved):

belforte:
I approved since things look good to me. The question on the code which I put may simply be a matter of educating me, and as long as you tested already with python3, go ahead and merge.

ddaina:
thanks. indeed, i tested this use case and it worked fine. also, please note that we have more code examples of such syntax, e.g. https://github.com/dmwm/CRABClient/blob/8cb30afa43ec0436356fa4162ad99f704017605f/bin/crab.py#L181

in any case, i will merge now and update crabclient in devel branch. if @mapellidario has suggestion how to change it, we can do it later

bin/crab.py
mapellidario: Doesn't this `exitcode` come from

https://github.com/dmwm/CRABClient/blob/e3741f8dfe02a39e972013f85d54866f6e34732f/src/python/CRABClient/ClientExceptions.py#L98-L102

?

If this is the case, then there is nothing to worry about, this is ok in both py2 and py3.

ddaina: yes, you are right

belforte: thanks Dario and Daina

PR #5123, opened 2021-11-30
add gocurl option

ddaina:
currently it does not work because of the issue with gocurl. There is something wrong with capath variable and trying to execute command ends up with bunch of messages like this:
```
[ddirmait@lxplus768 crab_tasks]$ /cvmfs/cms.cern.ch/cmsmon/gocurl -verbose 2 -method GET -header "User-Agent: CRABClient/development" -header "Accept: */*" -data "subresource=delegatedn" -cert "/tmp/x509up_u93401" -key "/tmp/x509up_u93401" -capath "/etc/grid-security/certificates/" -url "https://cmsweb-testbed.cern.ch:8443/crabserver/preprod/info?subresource=delegatedn" | tee /dev/stderr
<...>
2021/11/30 14:39:36 main.go:231: invalid PEM format while importing trust-chain: "/etc/grid-security/certificates//036b3363.namespaces"
2021/11/30 14:39:36 main.go:231: invalid PEM format while importing trust-chain: "/etc/grid-security/certificates//036b3363.r0"
<...>
```

need to open issue in https://github.com/vkuznet/gocurl

### DIFF
--- a/src/python/CRABClient/CrabRestInterface.py
+++ b/src/python/CRABClient/CrabRestInterface.py
@@ -174,14 +174,27 @@ class HTTPRequests(dict):
             url = url + '?' + data
 
         command = ''
-        command += 'curl -v -X {0}'.format(verb)
-        command += ' -H "User-Agent: %s/%s"' % (self['userAgent'], self['version'])
-        command += ' -H "Accept: */*"'
-        command += ' --data @%s' % path
-        command += ' --cert "%s"' % self['cert']
-        command += ' --key "%s"' % self['key']
-        command += ' --capath "%s"' % caCertPath
-        command += ' "%s" | tee /dev/stderr ' % url
+
+        # CRAB_useGoCurl env. variable is used to define how request should be executed
+        # If variable is set, then goCurl is used for command execution: https://github.com/vkuznet/gocurl
+        if os.getenv('CRAB_useGoCurl'):
+            command += '/cvmfs/cms.cern.ch/cmsmon/gocurl -verbose 2 -method {0}'.format(verb)
+            command += ' -header "User-Agent: %s/%s"' % (self['userAgent'], self['version'])
+            command += ' -header "Accept: */*"'
+            command += ' -data "%s"' % data
+            command += ' -cert "%s"' % self['cert']
+            command += ' -key "%s"' % self['key']
+            command += ' -capath "%s"' % caCertPath
+            command += ' -url "%s" | tee /dev/stderr ' % url
+        else:
+            command += 'curl -v -X {0}'.format(verb)
+            command += ' -H "User-Agent: %s/%s"' % (self['userAgent'], self['version'])
+            command += ' -H "Accept: */*"'
+            command += ' --data @%s' % path
+            command += ' --cert "%s"' % self['cert']
+            command += ' --key "%s"' % self['key']
+            command += ' --capath "%s"' % caCertPath
+            command += ' "%s" | tee /dev/stderr ' % url
 
         # retries this up at least 3 times, or up to self['retry'] times for range of exit codes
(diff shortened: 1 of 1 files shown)


belforte:
does it work if one does not specify capath ? I think we tested access to cmsweb from lxplus.

src/python/CRABClient/CrabRestInterface.py
```diff
@@ -174,14 +174,27 @@ def makeRequest(self, uri=None, data=None, verb='GET'):
...
-        command += 'curl -v -X {0}'.format(verb)
-        command += ' -H "User-Agent: %s/%s"' % (self['userAgent'], self['version'])
-        command += ' -H "Accept: */*"'
-        command += ' --data @%s' % path
-        command += ' --cert "%s"' % self['cert']
-        command += ' --key "%s"' % self['key']
-        command += ' --capath "%s"' % caCertPath
-        command += ' "%s" | tee /dev/stderr ' % url
+
+        # CRAB_useGoCurl env. variable is used to define how request should be executed
```
belforte, line 178: add one line comment like this:
# same variable is used also inside CRABServer, we should keep name changes (if any) synchronized

belforte (approved):
looks good. But indeed we should figure out what to do with capath before closing the issue

belforte:
in your test example at the top seems like there is a duplication:
 -data "subresource=delegatedn" 
and
-url "https://cmsweb-testbed.cern.ch:8443/crabserver/preprod/info?subresource=delegatedn"

src/python/CRABClient/CrabRestInterface.py
ddaina: thanks, done

belforte:
I am not sure about the "it does not work" at the top.
It looks to me that it spits a lot of error messages, but it works (at least works on lxplus). Somehow /etc/grid-security/certificates contains PEM files AND also a lot of other files which are likely useless/ignored by go http library and result in the long list of complains about "not PEM format".
Leaving aside for a moment the possible issue of filtering that noise out, and since on lxplus things work also w/o --capath, do you remember where we had the need to add it, so that we can check  there if the complains about non PEM files are harmful ?

belforte:
anyhow it should be easy for Valentin to silently skip non-PEM files instead of complaining here, sort of .. one needs to know what to expect.
https://github.com/vkuznet/gocurl/blob/47f0a476a1f6aec3fb69be54c315f07e149a3353/main.go#L218-L232

src/python/CRABClient/CrabRestInterface.py
```diff
@@ -174,14 +174,27 @@ def makeRequest(self, uri=None, data=None, verb='GET'):
...
+        # If variable is set, then goCurl is used for command execution: https://github.com/vkuznet/gocurl
+        if os.getenv('CRAB_useGoCurl'):
+            command += '/cvmfs/cms.cern.ch/cmsmon/gocurl -verbose 2 -method {0}'.format(verb)
+            command += ' -header "User-Agent: %s/%s"' % (self['userAgent'], self['version'])
+            command += ' -header "Accept: */*"'
+            command += ' -data "%s"' % data
```
ddaina, line 185: hm, we should also pass here file, not data.

belforte: I do not understand what you mean. If you want feedback, please elaborate. For GET there is no data, everything is in the URL. Do you refer to the file where we write the data for POST/PUT calls to avoid the problem with curl argument length ?

belforte:
about the spurious error messages from gocurl, I have opened https://github.com/vkuznet/gocurl/issues/5

ddaina:
please test

ddaina:
test this please

belforte:
test this please

PR #5129, opened 2021-12-08
cleanup tmp status-cache files. Fix #5128

### LINKED ISSUE #5128: crab client does not clean files in /tmp
Seems these 2 temporary files [1] are created but never removed. We just got users filling up /tmp (with more than 50GB of status-cache-...) in less than 1 day.
https://github.com/dmwm/CRABClient/blob/master/src/python/CRABClient/Commands/status.py#L115-L116

belforte:


belforte:
@ddaina thanks for pointing it out, it is more elegant this way

ddaina (approved):

belforte:
please @ddaina review again. Now all files created in /tmp should be removed

ddaina (approved):
looks good

PR #5136, opened 2021-12-10
warn users of crab-dev. Fix #5134

### LINKED ISSUE #5134: warn users of crab-dev
with ref to #5133 I'd do in two steps:
1. always print a gentle warning when using crab-dev
2. make a new version now to be put in crab-dev where there is more forceful message to "STOP" and remove it when it will be time to move this to production (alas it will need a new tag). So this is just a temporary thing to wean CMSSW_12 users out of the "use crab-dev" mantra of last few months

Step 1. can be done as simply as with this in `crab.py` (to be verified in actual deployment)
```python
if 'dev' in __file__ :
   print ("beware!")
````

belforte:


belforte:
@ddaina I do not see good ways to properly test this w/o deploying it in IB. Shall we make a new tag and do that ? (after a quick review of course). Or do you have some other idea ?
I tested a "variant" where I look for 'belforte' string in `__file__` and it worked. But the real test is that the path indeed always contain `crab-dev` string

ddaina (approved):
it looks good to me, we can deploy to IB.

PR #5137, opened 2021-12-10
pass bytes to b64encode. Fix #5127

### LINKED ISSUE #5127: bytes vs. string in crab killl by operator
I do not know if this is to be fixed on client or server side.
It is indeed an almost never used funcionality !!

```
belforte@lxplus7101/C12> crab kill -d  crab_20211202_104835 --killwarning 'killed by Stefano'
ERROR: TypeError: a bytes-like object is required, not 'str'

```

see also https://github.com/dmwm/CRABServer/issues/6896
once understood where problem is, close of these two identical tkts. Unless solution requires a change in both

belforte:


src/python/CRABClient/Commands/kill.py
```diff
@@ -26,7 +26,7 @@ def __call__(self):
         self.logger.debug("Killing task %s" % self.cachedinfo['RequestName'])
         inputs = {'workflow' : self.cachedinfo['RequestName']}
         if self.options.killwarning:
-            inputs.update({'killwarning' : b64encode(self.options.killwarning)})
+            inputs.update({'killwarning' : b64encode(self.options.killwarning.encode('utf-8'))})
```
mapellidario, line 29: This change will work with a specific set of assumptions, which may not be obvious at first:

- in py3: `self.options.killwarning` needs to be unicode. if it is not (maybe under some condition this is already a bytes string?), this will raise [1]

- in py2: `self.options.killwarning` needs to be either
  - unicode string. same consideration as py3 apply here
  - a bytes string, containing only ascii characters. This is because py2 tries to be smart and tries to automatically decode the bytes string to unicode with `ascii` codec before encoding it to utf8 as requested. if `self.options.killwarning` is a bytes string (which is the default py2 `str`) containing non-ascii characters, we have [2]

Since we want not to rely on dmwm/WMCore and since we want to keep py2 and py3 compatibility in our Client, I suggest that we create our own set of utilities for managing strings, in a similar way to what dmwm/WMCore does. In particular, I suggest that we create at least
- `encodeUnicodeToBytes`
- `decodeBytesToUnicode` 

In particular, I would use them here as

```diff
- inputs.update({'killwarning' : b64encode(self.options.killwarning.encode('utf-8'))})
+ inputs.update({'killwarning' : b64encode(encodeUnicodeToBytes(self.options.killwarning))}) 
```

This would fix all the aforementioned potential issues.

Since we do not want to rely on python-future, we need some juggling and define our functions with some `if`s:

```python
import sys
PY2 = sys.version_info[0] == 2
PY3 = sys.version_info[0] == 3 # or also >=3

if PY2:
    mybytes = str
    myunicode = unicode 
if PY3:
    mybytes = bytes
    myunicode = str

def encodeUnicodeToBytes(value):
    if isinstance(value, myunicode):
        return value.encode("utf-8", errors)
    return value

def decodeBytesToUnicode(value):
    if isinstance(value, mybytes):
        return value.decode("utf-8")
    return value
```

[1]: 

```plaintext
# python3
>>> b"hello".encode("utf8")
  File "<stdin>", line 1, in <module>
AttributeError: 'bytes' object has no attribute 'encode'
```

[2]: 

```plaintext
# python2
>>> "♄℮ℓℓø".encode("utf8")
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 0: ordinal not in range(128)
```

mapellidario (commented):
Good that we caught this error!

I propose to use our own version of `encodeUnicodeToBytes` and `decodeBytesToUnicode` to have a consistent overview of where we do this manipulation. 

It is likely that your `.encode()` will fix this issue in 99% of the times, but this is a good moment for stopping and planning. Do we want to deploy a quickfix, which we will benefit from immediately but that we risk to come back and fix again in the future in case of non-ascii arguments, or do we want to spend a little time creating two functions that hide the complexity of dealing with the proper string encoding/decoding based on the specific python version and that we will greatly benefit from in the future?

src/python/CRABClient/Commands/kill.py
belforte: all true. But this is the only instance of `b64encode` in CRABClient, and until/unless some operator gets very creating with what to type inside the quotes of
```
crab kill -d somedir --killwarning 'killed by me'
```
it is reasonable to expect that there will be no problems, and the day someone tries to put non-ascii in there and gets and error...

So this would be more like:
add those utilities to CRABClient anyhow, go through all places where py3 compatibility was obtained in a hacky or risky way, and do it more professionally. And of course the question is "is it worth". The hardest part at this point would be to find all those places (besides the obvious dozen of places where `sys.version_info` is currently called).
Should we open such a larger project and try to get an idea of the amount of work during next months, at low priority ?

belforte: note that this `--killwarning` option is only available to CRAB operators and I can't remember if we ever used it

mapellidario: > note that this --killwarning option is only available to CRAB operators and I can't remember if we ever used it

Aha ok, this defeats my argument for "sanitizing" user input. All clear! I think we can live with unicodedecoding errors to be presented to operators when they insert non-ascii characters!

>  Should we open such a larger project and try to get an idea of the amount of work during next months, at low priority ?

Yes! I am definitely in favor of such endeavor. I know it may not sound like superfun, but it is the only way to have some confidence and sleep well at night

belforte: Sounds good Dario. Thanks. I have created a new issue for this https://github.com/dmwm/CRABClient/issues/5139 feel free to change the description in there

belforte:
@mapellidario can you approve the PR then ? (just to do things by-the-book)

mapellidario (approved):

PR #5150, opened 2022-02-08
Include the correct twiki link for CRAB3FAQ

tvami:
Simple PR for fixing the fact that
https://twiki.cern.ch/twiki/bin/view/CMSPublic/CRAB3CommonErrors#Syntax_error_in_CRAB_configurati
leads to a twiki that says that every relevant info was moved to
https://twiki.cern.ch/twiki/bin/view/CMSPublic/CRAB3FAQ


belforte:
thanks @tvami for pointing out the broken link.

tvami:
So I guess this should have also been directed towards the `python3` branch, right?

belforte:
no, this was fine. Sorry for the confusion.
for CRAB Client we choose to make a separate branch with the old py2 code and use master to put in py3 compatibility. In the end we found it very difficult to keep the two aligned, and so decided to try the other way around for CRABServer repository.

tvami:
Ahh I see, ok thanks!

PR #5161, opened 2022-03-23
look for output modules in FinalPaths as well. Fix #5160

### LINKED ISSUE #5160: make CMSSWConfig find output files also in FinalPath
so far CRAB automatically finds EDM output files in output module `PoolOutputModule` defined as `cms.EndPath`, but now there's also the possibility to define output as `cms.FinalPath`.
See: https://github.com/cms-sw/cmssw/issues/37301
and in particular
https://github.com/cms-sw/cmssw/issues/37301#issuecomment-1075536022

belforte:


belforte:
tested running locally from GH

PR #5180, opened 2022-11-30
adapt CRABClient to new jobwrapper

mapellidario:
We need to update crab client `submit --dryrun` and `preparelocal` commands to match the new file structure introduced with https://github.com/dmwm/CRABServer/releases/tag/v3.221128 (new job wrapper)


belforte:
so we need to have this in production before we can update job wrapper in TW.

OUCH !

belforte:
and of course modify this PR to make it backward compatible !

mapellidario:
I added a second commit, which implements backwards compatibility with the old TW/jobwrapper.

I both `submit --dryrun` and `preparelocal` with this new code against both prod (old jobwrapper) and preprod (new jobwrapper) and everything works.

I think this PR is now ready to be review, merged, deployed, ...

belforte (approved):
Very neat handling of thebackward compatibility. Thanks Dario

PR #5185, opened 2023-02-07
new Data.inputBlocks parameter

novicecpp:
Fix https://github.com/dmwm/CRABServer/issues/7203

Client side of https://github.com/dmwm/CRABServer/pull/7525
`Data.inputBlocks` as list of input blocks that users want to processes. 

~~Currrently drafting. We can lets users supply only blocks uuid then we concat its with inputDataset in client side.~~
~~And/or lets user supply full block name but we need to verify if dataset's name in block is same as inputDataset~~

I decide to allow users to provide either full block name or block UUID. 
- If it match UUID format, client will append dataset name for users before sending to server.
- If not, let server report error in case they forget to change dataset name.


belforte:
Looks very good, thanks.
Hey... super nice to users !
> I decide to allow users to provide either full block name or block UUID.

belforte (approved):

PR #5188, opened 2023-02-17
add src/**/python to user sandbox

### LINKED ISSUE #5187: Add src/*/*/python to the user sandbox
### description

The change https://github.com/cms-sw/cmsdist/pull/7865 stopped creating symlinks that CRABClient relied upon when the parameter `JobType.sendPythonFolder` is set to `True` in user task config.

### proposed solution

Shahzad on CMSTalk suggested to always add `src/*/*/python` and `src/*/*/interface` to the user sandbox [1].

We already add `src/**/interface` https://github.com/dmwm/CRABClient/blob/b1d3d94c9ba5e2209f9687af21c34278e23804d8/src/python/CRABClient/JobType/UserTarball.py#L163

I suggest to simply add `python` to that line. A PR is on its way.

----

fyi: @rgerosa 

[1] https://cms-talk.web.cern.ch/t/missing-symbolic-links-in-cmssw-base-python/20403/35?u=dmapelli

mapellidario:
Fixes #5187

### status

tested:

- [x]  https://cmsweb-test11.cern.ch/crabserver/ui/task/230217_111711%3Admapelli_crab_20230217_121708

### description

This PR simply adds to the sandbox all the `python`  directories that are recursively found under `$CMSSW_BASE/src`. 

I have two questions for @belforte :

- [x] should we clean the comments and variable names? I mean, if we send the `python` directories maybe we should stop referring to those directories as "data". this is just cosmetic, i am not even sure i am willing to risk renaming variables and comments more than necessary
- [x] should we stop supporting the `Jobtype.sendPythonFolder` option? I am not sure how much this change makes it irrelevant. Maybe this is more for cmssw people to decide

### DIFF
--- a/src/python/CRABClient/JobType/UserTarball.py
+++ b/src/python/CRABClient/JobType/UserTarball.py
@@ -160,7 +160,7 @@ class UserTarball(object):
         # Note that dataDirs are only looked-for and added under the src/ folder.
         # /data/ subdirs contain data files needed by the code
         # /interface/ subdirs contain C++ header files needed e.g. by ROOT6
-        dataDirs = ['data', 'interface']
+        dataDirs = ['data', 'interface', 'python']
         userFiles = userFiles or []
 
         # Tar up whole directories
--- a/src/python/CRABClient/JobType/UserTarball.py
+++ b/src/python/CRABClient/JobType/UserTarball.py
@@ -145,6 +145,8 @@ class UserTarball(object):
         """
         Add the necessary files to the tarball
         """
+
+        # Tar up whole directories in $CMSSW_BASE/
         directories = ['lib', 'biglib', 'module']
         if getattr(self.config.JobType, 'sendPythonFolder', configParametersInfo['JobType.sendPythonFolder']['default']):
             directories.append('python')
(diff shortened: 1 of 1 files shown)


belforte (approved):

belforte:
sorry @mapellidario for revieweing so late. Feel free to merge and build new client

belforte:
I thought I had replied to questions above, looks like not. So:
> should we clean the comments and variable names?

  sure, thanks
> should we stop supporting the Jobtype.sendPythonFolder option?

 I think not, that's `$CMSSW_BASE/python` IIRC, while here you are adding `$CMSSW_BASE/src/<whatever>//python/`

mapellidario:
> I think not, that's $CMSSW_BASE/python IIRC, while here you are adding $CMSSW_BASE/src/<whatever>//python/

Yeah, you are right. `$CMSSW_BASE/python` will still be populated in all the cmssw version older than 12_6, so it makes sense to keep it. thanks!

I added some comments, my changes feel pretty safe, I will merge them myself and will run some tests on them.

Thanks for the review!

belforte:
AH !! I did not know that with CMSSW_12_6 they change the way user python scripts are managed ! Seems a large change for a mid-series version. Oh well... 
Honestly I did not know what exactly has to go in $CMSSW_BASE/python, we made it optional to include it simply in fear that users put all kind of junk there, but if people put there only what `scram` needs, we may be safe by removing the option and simplify things a bit.
If you get a chance to talk with Shahzad, maybe you can clarify ?

PR #5189, opened 2023-02-27
add Data.rucioScope

belforte:
if rucioScope is set in the configuration, add it to user_config dict so that TW will lookup dataset information in Rucio only


belforte:
I need to add some validation of the scope, especially submit should immediately fail if used w/o FileBased splitting

belforte:
the only error in pylint is this one which is totally harmless, but I do not know how to fix
I need that line `# pylint: disable=bad-whitespace` to avoid useless noise from pylint when testing locally
```
    src/python/CRABClient/ClientMapping.py
        E0012, line 3: Bad option value 'bad-whitespace' 
```

belforte:
On second thought, in line with what is done for other configuration parameters, validation and especially check of consistency among parameter values is done server side. I have added the check on splitting in TaskWorker. This PR is good as it for me. Please review.

belforte:
@mapellidario sorry to say that once this is merged it would be good to make a new CRABClient tag, but I think full validation does not need to be repeated.

mapellidario:
This looks good to me, despite I don't have much experience with client arguments. We could wait for wa's opinion, but it's also true that it feels pretty harmless. I am ok with you merging this by yourself!

In any case, tags are cheap! I can make a new one and add it to IB after we merge this PR. :)

belforte:
In the end we have a solution which requires no change to CRABClient !
Let's forget about this then.

PR #5192, opened 2023-04-03
add crab checkdataset. Fix #5191

### LINKED ISSUE #5191: implement crab checkdataset
should be based on
https://github.com/dmwm/CRABServer/blob/master/scripts/Utils/CheckDiskAvailability.py

plan:
modity crab checkwrite (which already talks to Rucio) to wget the script on the fly from GH and run it

belforte:


belforte:
tested in local GH clone

PR #5199, opened 2023-06-09
Add dirs to sandbox

belforte:


belforte:
tested, also with jenkins

PR #5207, opened 2023-09-01
skip .git directories when creating sandbox. Fix #5202

### LINKED ISSUE #5202: do not put `.git` directories in sandbox
see https://cms-talk.web.cern.ch/t/crab-submission-fails-tarball-issues/26625/2

Maybe exclude all hidden files/directories which start with `.` ? 

@smuzaffar  @makortel any advice from framework POV ? Are there hidden files in $CMSSW_BASE which we should be careful to keep ?

belforte:
includes a few small pylint fixes

### DIFF
--- a/src/python/CRABClient/JobType/UserTarball.py
+++ b/src/python/CRABClient/JobType/UserTarball.py
@@ -6,7 +6,6 @@
 
 from  __future__ import division   # make division work like in python3
 
-import json
 import os
 import glob
 import math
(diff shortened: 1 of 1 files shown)


belforte:
fixing remaining pylint messages would likely break py2 compatibility

mapellidario (approved):
Thanks Stefano, I like it! I am always amazed by the flexibility of the python standard library :)

(sorry for the late review!)

PR #5209, opened 2023-09-20
Print Rucio info for Rucio ASO task

novicecpp:
Before:
Screenshot from 2023-09-20 13-03-57

After:
Screenshot from 2023-09-20 12-58-51


belforte (approved):

belforte:
Looks good. Thanks

PR #5211, opened 2023-09-22
change getLumiListInValidFiles to use dasgoclient Fix #5175

### LINKED ISSUE #5175: update getLumiListInValidFiles() in UserUtilities
Current code only works in python2, and even in that case fails with new go-based DBS server.
See 
https://cms-talk.web.cern.ch/t/error-when-using-getlumilistinvalidfiles/16240/3

 - [ ] quick fix: make old client work with new server using the tip from Kifri [1]
 - [ ] add a warning in case it is called in python2 script.
 
long term fix: use dasgoclient - likely not worth !

[1] 
I edited the following lines:
https://github.com/dmwm/CRABClient/blob/7248a059fb2dba142c8633e52e788ba8095b1932/src/python/CRABClient/UserUtilities.py#L175-L177

to
```python
lumi = f['lumi_section_num']
runLumiPairs.append((run, lumi))
```
since `f['lumi_section_num']` is just an int. It now works as expected.

belforte:


belforte:
Test on normal dataset and on a dataset with invalid files.

PR #5214, opened 2023-10-06
intial implementation of adding Rucio client. Fix #5213

### LINKED ISSUE #5213: allow CRABClient to use Rucio python client API
see
https://github.com/cms-sw/cmsdist/issues/8737

Reporting conclusion here:

* CRABClient will use, on whichever OS it runs, the Rucio client for that arch as setup by 
```
source /cvmfs/cms.cern.ch/rucio/setup-py3.sh
```

* We will make sure that cmssw environment is not only changed during the execution of CRABClient commands (a forked subprocess in the end as per https://github.com/dmwm/CRABClient/blob/master/bin/crab

* We will restrict use of Rucio python client inside CRAB to CMSSW releases where it works, and too bad for people stuck in SL6 era.
  * initial very simple testing indicates that CMSSW_10_6_x is OK, so UL is covered, thanks Matti.
  * but for using that we need to switch to using python3 for CRABClient in there (currently there's a problem with pycurl)

belforte:


belforte:
@mapellidario I ended up in putting "all code" in the bash script `crab`
And added two small scripts to keep that terse.
But I am not particularly pleased. So any suggestion for a cleaner implementation is welcome.

belforte:
ready to test more

PR #5222, opened 2023-10-12
Initialise Rucio client after we have proxy. Fix #5220

### LINKED ISSUE #5220: only initialize Rucio when we have a proxy and if needed
current implementation effectively intializes Rucio client immediately
https://github.com/dmwm/CRABClient/blob/fae451b2c8e75614e73eba49aa05dd22cc9f936d/bin/crab.py#L30

Better to delay until we have ensure a valid proxy in SubCommand
https://github.com/dmwm/CRABClient/blob/fae451b2c8e75614e73eba49aa05dd22cc9f936d/src/python/CRABClient/Commands/SubCommand.py#L459-L461

and only if the command will require access to Rucio.

Rather than create yet another field in 
https://github.com/dmwm/CRABClient/blob/fae451b2c8e75614e73eba49aa05dd22cc9f936d/src/python/CRABClient/ClientMapping.py#L133
I will rename `initializeProxy` to `requiresRucio` since as indicated above ALL commands require a proxy !

belforte:


belforte:
tested on CMSSW_10, 11 and 13 in https://github.com/dmwm/CRABServer/issues/7943 https://github.com/dmwm/CRABServer/issues/7942 https://github.com/dmwm/CRABServer/issues/7941

PR #5225, opened 2023-10-16
make sure all commands return a dict. Fix #5224

### LINKED ISSUE #5224: all commands must return a dict containing `commandStatus` key
This is needed so that `crab.py` can know if a command succeeded of failed and properly set the exit code.
Currently commands always exit with 0 and rely on user reading the messages to tell if things worked.
This prevents constructing a decent testing.

Progress tracker

- [x] checkdataset
- [x]  checkusername
- [x]  checkwrite
- [x]  createmyproxy
- [x]  getlog
- [x]  getoutput
- [x]  kill
- [x]  preparelocal
- [x]   proceed
- [x]   remake
- [x]   report
- [x]   resubmit
- [x]  status
- [x]   submit
- [x]   tasks
- [x]   uploadlog

belforte:


belforte:
@mapellidario note that this assumes that https://github.com/dmwm/CRABClient/issues/5224#issuecomment-1765315793 is OK. I am still waiting yours and Wa's opinion on both the strategy and the  implementation.

mapellidario (approved):
I like it, thanks! 
I do not expect many users to use these return values, so i think it is pretty safe to change them and just announce the changes

src/python/CRABClient/Commands/remake.py
```diff
@@ -36,8 +37,15 @@ def remakecache(self,taskname):
...
-            return requestarea
 
+            caller = sys._getframe(1)
+            if not 'crab.py' in caller.f_code.co_filename:
+                msg = "ATTENTION return value for 'remake' has been changed to a dictionary"
+                msg += "\n format is {'commandStatus': 'SUCCESS' or 'FAILED',"
+                msg += "\n            'workDir': name of the work directory created}"
+                self.logger.warning(msg)
+            returnDict = {'commandStatus': 'SUCCESS', 'workDir': requestarea}
+            return returnDict
```
novicecpp, line 48: No need of warning message if you do not support backward compatibility.

bin/crab.py
```diff
@@ -142,7 +142,9 @@ def log_exception(exc_type, exc_value, tback):
             sys.exit(-1)
         self.cmd = sub_cmd(self.logger, args[1:])
 
-        self.cmd()
+        returnDict = self.cmd()
+        if returnDict['commandStatus'] != 'SUCCESS':
+            sys.exit(1)
```
novicecpp, line 147: We should return/raise back to `__main__` section instead of exit here.

novicecpp (commented):
I have some inline comments and also we should put some docstring about return value of subcommand (maybe in both SubCommand.py and RawCommand.py).

src/python/CRABClient/Commands/remake.py
belforte: I hope that a warning + a failure is more clear than just an obscure failure on some code which used to work. It surely makes no sense long term, but simply to help the transition. I do not think I have a way to catch if user code fails on my new return value or not.

bin/crab.py
belforte: I agree. I put `sys.exit(1)` becasue raising ClientException did not seem adequate. In a way what I want is to make them end with exit code 1 some commands that until now reports a "failure" in the messages but return with exit code 0.
I will define a new exception for this

novicecpp:
Yes.
Thanks Stefano.

belforte:
all those pylint errors are due to "old code which needs to run in python2 as well, did not follow naming conventions and is anyhow untouched and working'

PR #5241, opened 2023-10-25
Add change dbs dataset file status fix #5204

### LINKED ISSUE #5204: import DBS3SetDatasetStatus and DBS3SetFileStatus into CRAB
- [x] copy  files from https://github.com/dmwm/DBS/tree/master/Client/utils/DataOpsScripts to our repo
- [x] make them py2/3 compatible
- [x] make them work as `crab` commands. We should replace calls to new DBS3 python client with direct `curl` to server (as per Valentin's instructions) so that code will work with both py2 and py3

- [ ] Also update instructions in https://twiki.cern.ch/twiki/bin/view/CMSPublic/Crab3DataHandling#Changing_a_dataset_or_file_statu
- [ ] Add to Jenkins ClientValidation test

belforte:
I think this is OK now. Aside from
* setfilestatus does not work with a list of files. Valentin acknowledged a bug but it is for others to fix. We can live w/o it for quite some time
* tune/polish help (remove single char commands ? now that we have autocompletion...)
* whatever @novicecpp and @mapellidario suggest


belforte:
and of  course I need to test on python2 and make sure I did not break anything !

src/python/CRABClient/Commands/setdatasetstatus.py
```diff
@@ -0,0 +1,167 @@
...
+
+def getDbsREST(instance=None, logger=None, cert=None, key=None, version=None):
+    """
+    given a DBS istance (e.g. prod/phys03) returns a DBSReader and DBSWriter
+    client instances which communicate with DBS REST via curl
+    Arguments:
+    logger: a logger
+    cert, key : name of files, can use the path to X509_USER_PROXY for both
+    version: the CRAB Client version to put in the User Agent field of the query
+    """
```
novicecpp, line 33: Why do you need to explicitly set CRABClient version here instead of read from usual module version?

src/python/CRABClient/CrabRestInterface.py
```diff
@@ -182,6 +184,8 @@ def makeRequest(self, uri=None, data=None, verb='GET'):
             command += '/cvmfs/cms.cern.ch/cmsmon/gocurl -verbose 2 -method {0}'.format(verb)
             command += ' -header "User-Agent: %s/%s"' % (self['userAgent'], self['version'])
             command += ' -header "Accept: */*"'
+            if self['Content-type']:
+                command += ' -header "Content-type: %s"' % self['Content-type']
             if verb in ['POST', 'PUT']:
                 command += ' -header "Content-Type: application/x-www-form-urlencoded"'
```
novicecpp, line 190:
```suggestion
            if self['Content-type']:
                command += ' -header "Content-type: %s"' % self['Content-type']
            else:
                command += ' -header "Content-Type: application/x-www-form-urlencoded"'
```
When body (`--data`) is supplied in curl, default `Content-Type` is `application/x-www-form-urlencoded`.

novicecpp (changes requested):

src/python/CRABClient/Commands/setdatasetstatus.py
belforte: VERY good question ! I suspect for no good reason, I will address toghether with "move to ClientUtilities"

src/python/CRABClient/Commands/setdatasetstatus.py
```diff
@@ -0,0 +1,167 @@
...
+    logger.debug('Write Url = %s' % dbsWriteUrl)
+
+    dbsReader = HTTPRequests(hostname=dbsReadUrl, localcert=cert, localkey=key,
+                             retry=2, logger=logger, verbose=False, contentType='application/json',
+                             userAgent='CRABClient', version=version)
+
+    dbsWriter = HTTPRequests(hostname=dbsWriteUrl, localcert=cert, localkey=key,
+                             retry=2, logger=logger, verbose=False, contentType='application/json',
+                             userAgent='CRABClient', version=version)
+    return dbsReader, dbsWriter
```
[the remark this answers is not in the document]
belforte: No good reasons. I had this in mind already. Should also change a bit HTTPRequests rename that stupid `uri` arguments as a saner `api` but that requires changing in a lot of places. Anyhow, I will move it

src/python/CRABClient/Commands/setfilestatus.py
```diff
@@ -0,0 +1,156 @@
...
+        if dataset:
+            data = {'dataset': dataset, 'is_file_valid': statusToSet}
+        jdata = json.dumps(data)  # PUT requires data in JSON format
+        out, rc, msg = dbsWriter.put(uri='files', data=jdata)
+        if rc == 200 and msg == 'OK':
+            self.logger.info("File(s) status changed successfully")
+            result = 'SUCCESS'
+        else:
+            msg = "File(s) status change failed: %s" % out
+            raise CommandFailedException(msg)
```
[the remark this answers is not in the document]
belforte: This would have better fit #5225 !
The only reasons I had to use both is that e.g. users of `status` API are used to always get a dictionary back, so I do not want to suddenly raise exceptions. In other cases, esp. where commands was already capable of rising exceptions, I thought that raising was making for cleaner code. Fewer "if".

src/python/CRABClient/CrabRestInterface.py
belforte: those lines are under the `useGoCurl` which is 1) little tested 2) not used atm 3) . So I did not pay much attention, sorry. Nor tested (bad, I know). I am not  sure that the difference eventually is since  `default Content default Content-Type is application/x-www-form-urlencoded` seems to apply to all verbs, IIUC. So in a way the original lines were likely useless ? Showing master branch here:
https://github.com/dmwm/CRABClient/blob/0bc1ad01ad739a1c6219a40c0cbd4882bc1ed3f4/src/python/CRABClient/CrabRestInterface.py#L185-L186

Shouldn't I then use same suggested change as above for the code that follows the `else:` and applies when not using GoCurl ?

belforte:
thanks a lot @novicecpp 
I have a question for you now: would it then be better to also move HTTPRequests to ClientUtilities ? Since it is not used by CRABRest only.

belforte:
I merged [my branch](https://github.com/belforte/CRABClient/tree/add-change-DBS-dataset-file-status-fix-5204) with master dmwm repo  in [test-setdatasetfilesstatus branch](https://github.com/dmwm/CRABClient/tree/test-setdatasetfilesstatus) tagged as [testSetDsetFilesStatus.1](https://github.com/dmwm/CRABClient/releases/tag/testSetDsetFilesStatus.1) and ran Jenkins CV to check for regression
* https://github.com/dmwm/CRABServer/issues/7977
* https://github.com/dmwm/CRABServer/issues/7978
* https://github.com/dmwm/CRABServer/issues/7979
* https://github.com/dmwm/CRABServer/issues/7980
* https://github.com/dmwm/CRABServer/issues/7981

All OK !

belforte:
@novicecpp I captured your comments before in [this](https://github.com/dmwm/CRABClient/issues/5204#issuecomment-1781051917) list of actions. I will add more as needed once you answer my questions. Feel free to check and comment in there. Thanks

src/python/CRABClient/CrabRestInterface.py
novicecpp: No need. Curl is working fine because it is [default behavior from curl when you supply `--data` param](https://everything.curl.dev/http/post/content-type), also apply to GET I tested in my local machine. However I am not sure about gocurl, so I suggested as above.

src/python/CRABClient/Commands/setfilestatus.py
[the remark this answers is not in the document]
novicecpp: I do not have any suggestion for this. I just want to know and never notice when I reviewed https://github.com/dmwm/CRABClient/pull/5225.
Thanks for explanation.

novicecpp:
> I have a question for you now: would it then be better to also move HTTPRequests to ClientUtilities ? Since it is not used by CRABRest only.

Maybe move HTTPRequests to utils and move DBSREST to new file same level as CrabRestInterface.py?

src/python/CRABClient/CrabRestInterface.py
belforte: OK. Let's do same way for `GoCurl` and `curl`. If we ever come to wanting to use `GoCurl` we will see. Most likely Valentin did things correctly and behaves exactly same as `curl`.

belforte:
Hope I addressed all coments (only the `uri` vs. `api` confusion is left, but I think it is too much work). Will run Jenkins again.

belforte:
code with last changes is now tagged as [testSetDsetFilesStatus.2](https://github.com/dmwm/CRABClient/releases/tag/testSetDsetFilesStatus.2) in the ad-hoc test branch

belforte:
ooopppss. I had added content-type in the wrong place. But of course CRAB Server REST happily ignores it and everything kept working, but better not confuse the reader.

belforte:
If there are no more objection I will do a BIG SQUASH and merge.

src/python/CRABClient/Commands/setdatasetstatus.py
```diff
@@ -0,0 +1,118 @@
...
+
+if sys.version_info >= (3, 0):
+    from urllib.parse import urlencode  # pylint: disable=E0611
+if sys.version_info < (3, 0):
+    from urllib import urlencode
+
+
+class setdatasetstatus(SubCommand):
+    """
+    Set status of a USER dataet in phys03,
```
mapellidario, line 22: small typo here

src/python/CRABClient/Commands/setdatasetstatus.py
```diff
@@ -0,0 +1,118 @@
...
+    optionally invalidates/revalidates all files in it
```
mapellidario, line 23: I have a question: does this `optionally` means that it is possible to have valid files in an invalid dataset?
I mean, i have a valid dataset with valid files. When happens when i call `setdatasetstatus` without the `--recursive` argument? only the dataset is invalidated, or the dataset and all the files in it?

src/python/CRABClient/Commands/setdatasetstatus.py
```diff
@@ -0,0 +1,118 @@
...
+                               help="DBS instance. e.g. prod/phys03 (default) or int/phys03. Use at your own risk." + \
+                                    "Unless you really know what you are doing, stay with the default"
+                               )
+        self.parser.add_option('--dataset', dest='dataset', default=None,
+                               help='dataset name')
+        self.parser.add_option('--status', dest='status', default=None,
+                               help="New status of the dataset: VALID/INVALID/DELETED/DEPRECATED",
+                               choices=['VALID', 'INVALID', 'DELETED', 'DEPRECATED']
+                               )
+        self.parser.add_option('--recursive', dest='recursive', default=False, action="store_true",
```
mapellidario, line 93: just to make sure i understand correctly, this applies to children datasets and their files, but not to the files of the current dataset that is being invalidated, doesn't it?

src/python/CRABClient/Commands/setfilestatus.py
```diff
@@ -0,0 +1,150 @@
...
+
+from CRABClient.Commands.SubCommand import SubCommand
+from CRABClient.ClientExceptions import MissingOptionException, ConfigurationException, CommandFailedException
+from CRABClient.ClientUtilities import colors
+from CRABClient.RestInterfaces import getDbsREST
+
+
+class setfilestatus(SubCommand):
+    """
+    Set status of a USER dataset in phys03,
```
mapellidario, line 16: this command does not touch dataset status, right?

mapellidario (changes requested):

src/python/CRABClient/Commands/setdatasetstatus.py
belforte: oops :-) fixed,

novicecpp (approved):
Looks good now. Thanks.

src/python/CRABClient/Commands/setfilestatus.py
belforte: right

src/python/CRABClient/Commands/setdatasetstatus.py
belforte: the idea (from the old DBS script) is to also invalidate all files in current dataset, no matter if --recursive is there or not :man_shrugging:

src/python/CRABClient/Commands/setdatasetstatus.py
belforte: dataset status and file status are independent concepts iin DBS.
Files only have a boolean flag: is_file_valid=0/1
In practice this is used to mark files lost or otherwise unreadable.

Dataset status is a flag which can have many values and it is used by P&R also to track status, e.g. PRODUCTION means the dataset is sitll being populated. Other important values are:
DEPRECATED (e.g. there is newer version, but it could still be used e.g. to compare)
INVALID (contents are known to be no good for physics, bugs in the code, bad conditions.. but again there can be reasons to look at data)
DELETED (the name says it all)
In CRAB we always told users to only use VALID/INVALID for datasets, to avoid creating confusion. In the end if you flag your USER dataset as INVALID, nobody should try to read w/o checking with you.

So in a way is_file_valid is only meaningful for non-DELETED datasets, and its function is to allow data processors to skip missing files. We never delete records from DBS and we never delete the info that a file was part of the dataset at some time, simply record that it became unreadable (and should it happend that it gets rescued, we can still turn it back into valid).

In theory that's all and there is no need to invalidate all files in a dataset.
In practive, few people care to know the full story, and it is not written anywhere, and P&R may change the meaning of the dataset status as they please, so it makes some sense that if you delete your USER dataset from disk, you also want to mark all files with  `is_file_valid=0`, just to be on the safe side.

By default the files status is not affected by set dataset status.
And the idea for --recursive is to flatly clean everything by marking as invalid all files in the dataset.
It is not likely that users have children datasets to worry about. You may argue that you want to delete the parent and keep childs around, so after the --recursive invalidation there will be a need for a few validation.

Note that I simply tried to reproduce the semantic of old scripts which were developed (long time ago) for production operators in which case marking children datasets as e.g. INVALID for important. I am not even sure that at that time operators really knew the difference between DELETED and INVALID. As I said nothing was ever written down and the fact that dataset validity is a string, and file validity one bit, makes it difficult to have consistency w/o strong conventions.  Ref: https://github.com/dmwm/DBS/blob/master/Client/utils/DataOpsScripts/DBS3SetDatasetStatus.py
Indeed I failed ! Since the original [would mark](https://github.com/dmwm/DBS/blob/14df8bbe8ee8f874fe423399b18afef911fe78c7/Client/utils/DataOpsScripts/DBS3SetDatasetStatus.py#L59) all files in a dataset as invalid if dataset type is no VALID. IMO that's bad. And I hope operators have not used this exact script, in the end (indeed only CRAB users complained when it stopped working).

That said, since --recursive is not implemented yet, we can do what we want. I am personally in favor of doing w/o it. Indeed the above scenario is a good part of why I decided to postpone it !

We can  clarify in the instructions that it is up to users to mark all files as well (pretty easy anyhow) and look for childrens (we can also put instructions e.g. [1]). In the end this `--recursive` in the original was confusing
and clearly ad hoc for some specific use case. 

Bottom lines, users should do:
1. I lost a file on disk: crab setfilestuatus  --file ... --status INVALID (once for every file until dbs2go bug is fixed)
2. I want to make sure nobody tries to read this (e.g. gone from disk):
   * crab setdataset --dataset ... --status INVALID (or DELETED)
   * AND TO BE FULLY SURE (since you can ask CRAB e.g. to read a dataset in any status):
   * crab setfilestatus --dataset ... --status INVALID



[1]
```
 belforte@lxplus805/TC3> curl --cert $X509_USER_PROXY  --key $X509_USER_PROXY -H "Accept: application/json" "https://cmsweb.cern.ch:8443/dbs/prod/global/DBSReader/datasetchildren?dataset=/ZeroBias5/Run2023F-v1/RAW"
[
{"child_dataset":"/ZeroBias5/Run2023F-PromptReco-v1/DQMIO","child_dataset_id":14767495,"dataset":"/ZeroBias5/Run2023F-v1/RAW"}
,{"child_dataset":"/ZeroBias5/Run2023F-LogErrorMonitor-PromptReco-v1/USER","child_dataset_id":14767497,"dataset":"/ZeroBias5/Run2023F-v1/RAW"}
,{"child_dataset":"/ZeroBias5/Run2023F-LogError-PromptReco-v1/RAW-RECO","child_dataset_id":14767500,"dataset":"/ZeroBias5/Run2023F-v1/RAW"}
,{"child_dataset":"/ZeroBias5/Run2023F-HcalCalIsolatedBunchSelector-PromptReco-v1/ALCARECO","child_dataset_id":14767504,"dataset":"/ZeroBias5/Run2023F-v1/RAW"}
,{"child_dataset":"/ZeroBias5/Run2023F-SiStripCalMinBias-PromptReco-v1/ALCARECO","child_dataset_id":14767508,"dataset":"/ZeroBias5/Run2023F-v1/RAW"}
,{"child_dataset":"/ZeroBias5/Run2023F-SiStripCalZeroBias-PromptReco-v1/ALCARECO","child_dataset_id":14767509,"dataset":"/ZeroBias5/Run2023F-v1/RAW"}
,{"child_dataset":"/ZeroBias5/Run2023F-TkAlMinBias-PromptReco-v1/ALCARECO","child_dataset_id":14767519,"dataset":"/ZeroBias5/Run2023F-v1/RAW"}
,{"child_dataset":"/ZeroBias5/Run2023F-PromptReco-v1/NANOAOD","child_dataset_id":14767525,"dataset":"/ZeroBias5/Run2023F-v1/RAW"}
,{"child_dataset":"/ZeroBias5/Run2023F-PromptReco-v1/MINIAOD","child_dataset_id":14767544,"dataset":"/ZeroBias5/Run2023F-v1/RAW"}
,{"child_dataset":"/ZeroBias5/Run2023F-PromptReco-v1/AOD","child_dataset_id":14767549,"dataset":"/ZeroBias5/Run2023F-v1/RAW"}
]
belforte@lxplus805/TC3> 
```

belforte:
thanks @mapellidario and @novicecpp for review and advice. I will merge. And we can discuss separately if/how to add --recursive. Maybe wait for someone to ask for it :grin: ?

We still have time to decide if to keep old semantic (Invalidating a dataset will automatically invalidate all the files in the dataset) or leave it to user with two actions as I indicated above. It seems a soft decisions, with arguments in pro/con either choice.

belforte:
Since in current implementation `crab setdatasetstatus` does not change status of individual files, I will print a note about this when the command is executed !

mapellidario:
Thanks Stefano for the detailed description of dataset and file validity in DBS. Sorry for the dumb questions! 

Do we already have an entry in the FAQs about DBS dataset and file status? Since we are allowing users to change it, we should also provide clear instructions about what those are and what the commands we provide do (or plan to do).

belforte:
Dario,your questions were not dumb. The problem is that there is no real rule on how to use those status flags,
DBS allows to set, then everybody can do as they see fit. I think Production changed their practice over time.

As user documentation  this is what we have:
  https://twiki.cern.ch/twiki/bin/view/CMSPublic/CRAB3FAQ#Can_I_delete_a_dataset_I_publish
it clearly needs a rewriting ! That is my next step.

PR #5242, opened 2023-10-25
avoid rucio setup getting confused by existing CMSSW stuff

belforte:
if one has CMSSW setup for a different arch than the current machine (e.g. do cmsrel on SL7 and than run on SL8, which works) the `/cvmfs/cms.cern.ch/rucio/setup-py3.sh` will get confused and fail to resolve dependencies.
This PR makes sure that Rucio is setup in the pre-cmsenv environment.


belforte:
well. this can surely only help and it is harmless. Let's merge so it will be tested by Jenkins

PR #5247, opened 2023-10-31
Use isEnoughRucioQuota from ServerUtilities.py

novicecpp:
Together with CRABServer PR https://github.com/dmwm/CRABServer/pull/7987

Use isEnoughRucioQuota from CRABServer's `ServerUtilities.py`, minimum CRABServer version is [v3.231103](https://github.com/dmwm/CRABServer/releases/tag/v3.231103).

~~Note for one big change; I have moved initialize Rucio client to SubCommand class method because we need to get proper Rucio account to get account usage. You can only allow to get usage from your own account unless you have admin privilege. The problem occur when we want to use group area where it has different account.~~

~~In example, users want store outputs to Rucio group area which need different Rucio account (In our case, we need `crab_test_group` account, not `belforte` or `dmapelli` account). To get the account, we parse the LFN by using `getRucioAccountFromLFN()` from ServerUtilities.py.~~

~~But, LFN information for some command (e.g., `status`) are available inside `__call__` method. [Here an example](https://github.com/novicecpp/CRABClient/blob/38517e8e10f884c9f732736dca3aa6e3c88efbe7/src/python/CRABClient/Commands/status.py#L68).~~

EDITED:

In the end (after reviewew), I have decided to stick with `self.cmdconf['requiresRucio']=True` feature because, in theory, you can get quota from any user, but there is an permission issue in Rucio server so workaround code has been put in place (see comment in https://github.com/dmwm/CRABClient/pull/5247#issuecomment-1825785879).

### DIFF
--- a/src/python/CRABClient/ClientExceptions.py
+++ b/src/python/CRABClient/ClientExceptions.py
@@ -101,3 +101,9 @@ class RESTInterfaceException(ClientException):
     Errors coming from interaction with REST interface
     """
     exitcode = 3016
+
+class RucioClientException(ClientException):
+    """
+    Errors coming from interaction with REST interface
+    """
+    exitcode = 3017
--- a/src/python/CRABClient/ClientUtilities.py
+++ b/src/python/CRABClient/ClientUtilities.py
@@ -797,3 +797,32 @@ def execute_command(command=None, logger=None, timeout=None, redirect=True):
         logger.debug('output : %s\n error: %s\n retcode : %s' % (stdout, stderr, rc))
 
     return stdout, stderr, rc
+
+def getRucioClientFromLFN(origClient, lfn, logger):
+    """
+    Get appropriate Rucio client with account parsing from LFN.
+
+    :param origClient:
+    :type origClient:
+    :param lfn: LFN
+    :type lfn: str
+    :param logger: logger object
+    :type logger: logging
+
+    :return: rucio client object
+    :rtype: rucio.client.Client
+    """
+    from ServerUtilities import getRucioAccountFromLFN
+    from rucio.client import Client
+    from rucio.common.exception import RucioException
+    account = getRucioAccountFromLFN(lfn)
+    if origClient.account == account:
+        return origClient
+    try:
+        client = Client(account=account)
+        me = client.whoami()
+        logger.debug('Initializing new Rucio client for account %s' % me['account'])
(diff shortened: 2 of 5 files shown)


belforte:
I can see why you did this, but in view of introducing more commands with `self.cmdconf['requiresRucio']=True` I would like to keep Rucio client initialization in SubCommand, like before this PR. And inside specific commands, where a different account is needed, and where the LFN is known, we can switch account name in the client, even possibly restoring it if we want to make it possible to combine commands in a single script via the python API. 
What do you think ?

novicecpp:
Rebase wrong branch :man_facepalming: 
Sorry for noise.

belforte:
also note this very odd output. Rucio says that
 it failed, but prints quotas anyhow.
We need to know what exactly to expect here.
```
belforte@lxplus805/~> rucio whoami
email      : Stefano.Belforte@cern.ch
deleted_at : None
updated_at : 2020-04-28T23:25:20
suspended_at : None
account_type : USER
account    : belforte
status     : ACTIVE
created_at : 2020-04-28T23:25:20
belforte@lxplus805/~> rucio list-account-usage crab_test_group
+----------------+------------+------------+--------------+
| RSE            | USAGE      | LIMIT      | QUOTA LEFT   |
|----------------+------------+------------+--------------|
| T1_DE_KIT_Disk | 863.937 MB | 500.000 GB | 499.136 GB   |
| T2_IT_Legnaro  | 0.000 B    | 2.000 TB   | 2.000 TB     |
+----------------+------------+------------+--------------+
2023-11-20 12:46:29,507	ERROR	Access to the requested resource denied.
Details: Account belforte can not list global account usage.
belforte@lxplus805/~> 
```

novicecpp:
Well, it is indeed the problem we discussed with Eric in DM Dev channel in MM.
But for this PR and the fact that you can probe your own quota stats by default, I initialize a new client with the proper account from parsing LFN, so it works with both group accounts and as an API.

belforte (approved):

belforte:
Very nice, thanks.

belforte:
we should put here the minimum needed CRABServer tag to have the correct ServerUtilities

novicecpp:
[v3.231103](https://github.com/dmwm/CRABServer/releases/tag/v3.231103), also updated in the issue's description.

PR #5256, opened 2023-11-27
fix bug in #5254. Why didn't pylint spot this ?

belforte:


novicecpp:
https://github.com/dmwm/CRABServer/issues/7799#issuecomment-1668011801 :)

belforte:
thanks @novicecpp I will keep it in mind that variables in `self` are not checked

PR #5260, opened 2023-11-27
Move checkdataset to crab client repo fix 5233

belforte:
also introduce `commandUsedInsideCrab()` function in `ClientUtilities` so that we can decide if/when to use `print()` in a command for extra-verbosity of "progress report in same line" which only applies when using `crab` command, not the python Client API

PR #5262, opened 2023-11-29
make all file compile on py2/3 fix #5261

### LINKED ISSUE #5261: Code must be python2 compatible even if not used with py2
I broke that lately since I did not understand how python works 😭

see https://github.com/dmwm/CRABServer/issues/8025#issuecomment-1830536045

belforte:
some files (checkdataset.py, tasks.py) were taken care of in #5260


belforte:
I merge this so that I can test with Jenkins

PR #5287, opened 2024-01-30
preparelocal creates one json file with args for job. Fix #5276

### LINKED ISSUE #5276: preparelocal fails when `scriptArgs` is present
ref: https://cms-talk.web.cern.ch/t/issue-with-gfal-in-crab-jobs/32718/3

belforte:
also a small change to the Hellp message for `--instance`


belforte:
pylint complaints are the expected ones for py2 compatibility.

PR #5301, opened 2024-04-03
do not put objs-* in sandbox. Fix #5300

### LINKED ISSUE #5300: CRAB incorrectly following symlink in sandbox creation
Linktime optimization is a new speedup enabled in CMSSW_13_0_X and higher. It adds a directory `$CMSSW_BASE/external/${SCRAM_ARCH}/objs-base`, which is a symbolic link to `$CMSSW_RELEASE_BASE/objs/${SCRAM_ARCH}`. CRAB is following the symbolic link when creating the sandbox and adding all the files in that directory, which can exceed the allowed tarball size of 120 MB. Instead, the symlink should be preserved to avoid duplicating these contents from the release base.

belforte:


belforte:
tests in jenkins from master branch
 
   - [ ]  https://github.com/dmwm/CRABServer/issues/8313
   - [ ]  https://github.com/dmwm/CRABServer/issues/8314

belforte:
tested in https://github.com/dmwm/CRABServer/issues/8318

PR #5315, opened 2024-05-14
completions - generate_completions script works in py3

### LINKED ISSUE #5314: crab client bash completions do not respect custom sorting order
### problem

When using the crab client the suggestions do not respect the desired sorting. I get

```bash
[dmapelli@lxplus9105 ~]$ crab <tab><tab>
--debug           -h                checkwrite        kill              report            status
--help            checkdataset      createmyproxy     preparelocal      resubmit          submit
--quiet           checkfile         getlog            proceed           setdatasetstatus  tasks
--version         checkusername     getoutput         remake            setfilestatus     uploadlog
```

Despite we explicitly asked for

```bash
COMPREPLY=( $(compgen -W '--version --help -h --quiet --debug status tasks proceed checkwrite getlog checkusername checkdataset checkfile submit getoutput resubmit kill uploadlog remake report preparelocal createmyproxy setdatasetstatus setfilestatus' -- $cur) )
```

### solution

PR on its way...

mapellidario:
fixes #5314 

### status

tested on my laptop, but i have a small question before merging

### solution

This PR serves multiple purposes:

- make bash autocompletion respect the order at which we provide the commands. see the use of `-o nosort` in `complete -F _UseCrab -o filenames -o nosort crab`
- do not edit the `etc/crab-bashpcimpletion.sh` directly, generate it programmatically from `scripts/generate-completions.py`
  - make `generate-completions.py` work with python3
    - use `importlib` instead of  `imp`
  - make `generate-completions.py` generate the completion in our desired ordering, see the use of the `weights` variable.

open questions:

- [ ] I did not replicate to the variable `weights` the exact sorting that we have now yet. 
  - take a look at `etc/crab-bash-completion.sh` line 26. the order here is what bash shows when you press `crab <tab><tab>` (the order of line 32 will be the same as line 26)
  - what is the exact order that we want here?

### comment 

I decided to revamp the `generate_completion.py` script because I did not want to think about adding manually all the new commands introduced with `crab recover` :)

### DIFF
--- a/etc/crab-bash-completion.sh
+++ b/etc/crab-bash-completion.sh
@@ -23,101 +23,41 @@ _UseCrab ()
         "")
             case "$cur" in
             "")
-                COMPREPLY=( $(compgen -W '--version --help -h --quiet --debug status tasks proceed checkwrite getlog checkusername checkdataset checkfile submit getoutput resubmit kill uploadlog remake report preparelocal createmyproxy setdatasetstatus setfilestatus' -- $cur) )
+                COMPREPLY=( $(compgen -W '--version --help -h --quiet --debug status tasks submit proceed checkdataset checkfile checkusername checkwrite createmyproxy getlog getoutput getsandbox kill preparelocal recover remake report resubmit setdatasetstatus setfilestatus uploadlog' -- $cur) )
                 ;;
             -*)
                 COMPREPLY=( $(compgen -W '--version --help -h --quiet --debug' -- $cur) )
                 ;;
             *)
-                COMPREPLY=( $(compgen -W 'status tasks proceed checkwrite getlog checkusername checkdataset checkfile submit getoutput resubmit kill uploadlog remake report preparelocal createmyproxy setdatasetstatus setfilestatus' -- $cur) )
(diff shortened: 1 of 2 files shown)


belforte:
thanks Dario. I am not able to review this. I never knew that `generate_completion.py` existed and have no idea how and when to use it. Can you simply describe how things will work ?
Do we need to run it by hand every time we add or remove a command and pipe output to etc/crab-bash-completion.sh ?
Do you intend to have GH do this at any PR or tag ?

belforte:
And for sure I do not understand the code inside those scripts !

mapellidario:
I added a commit with and improved comment that i hope is clear enough.

So, at the moment I think we can use it instead of manually editing the file, which means run this script "only" when we add a crab client command or change an option in an existing command. 

In the future we can integrate this as part of the crab client build process, so that the completions are generated for every new deployment. I do not think the effort is worth at the moment.

mapellidario:
I addressed the some comments from pylint (the ones that i could quickly get rid of without a proper and useless refactoring). If everything goes well, I will merge this PR

mapellidario:
jenkins looks good to me, i will merge

PR #5320, opened 2024-06-28
Fix typos in output messages and subsection TWiki link for cmsenv

oljemark:
Hello. I found some typos in python files (non-frozen). Also changed the on-page paragraph link for the cmsenv setup section that had lost its target.


belforte (approved):

belforte:
thanks so much @oljemark for such a thorough review !

PR #5322, opened 2024-07-04
Fix autocompletion

### LINKED ISSUE #5319: add instructions to crab-completion.sh
see
https://github.com/cms-sw/cmsdist/pull/9261#issuecomment-2184074349

From Shahzad

having a comment like [a] just before the complete -F command in crab-bash-completion.sh will help for future updates

[a]
```text
######################################################
# Please make sure that 'complete` command below meet following rules
# - Starts with: complete\s+-F\s+<crab-function-name>\s+
# - Ends with: \s+crab
#  Otherwise suggest changes in crab-build.file in cms-sw/cmsdist repository
#####################################################
```

### LINKED ISSUE #5321: avoid `nosort` option in crab command completion
see
https://cms-talk.web.cern.ch/t/crab-is-not-working/41539/12?u=belforte

IMHO there is no need for `nosort` and we can also remove all those weights, listing commands in alphabetical order is fine. If we want to group command and/or add a tiny online guide, we can do in `crab --help` not in `crab <TAB>`

belforte:
fixes #5321 
fixes #5319

### DIFF
--- a/etc/crab-bash-completion.sh
+++ b/etc/crab-bash-completion.sh
@@ -23,13 +23,13 @@ _UseCrab ()
         "")
             case "$cur" in
             "")
-                COMPREPLY=( $(compgen -W '--version --help -h --quiet --debug status tasks submit proceed checkdataset checkfile checkusername checkwrite createmyproxy getlog getoutput getsandbox kill preparelocal recover remake report resubmit setdatasetstatus setfilestatus uploadlog' -- $cur) )
+                COMPREPLY=( $(compgen -W '--version --help -h --quiet --debug checkdataset checkfile checkusername checkwrite createmyproxy getlog getoutput getsandbox kill preparelocal proceed recover remake report resubmit setdatasetstatus setfilestatus status submit tasks uploadlog' -- $cur) )
                 ;;
             -*)
                 COMPREPLY=( $(compgen -W '--version --help -h --quiet --debug' -- $cur) )
                 ;;
             *)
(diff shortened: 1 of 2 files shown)


mapellidario:
Stefano, can I ask you to run the `generate_completions` script and commit the new ` etc/crab-bash-completion.sh` as well? [1]

maybe we should add this command to the specifile as well https://github.com/cms-sw/cmsdist/blob/IB/CMSSW_14_1_X/master/crab-build.file, but if the changes are not frequent enough it may not be worth it.

---

[1]

```bash
> cd CRABClient
> python3 scripts/generate_completions.py
```

mentioned at https://github.com/dmwm/CRABClient/blob/d17ff66a143d33d99d31c8f767c3cc75fdbceef8/scripts/generate_completion.py#L13

mapellidario (changes requested):
(changes suggested in previous comment)

belforte:
I had tried that, but can't do it on my laptop since
```text
LapSB:CRABClient$ python3 scripts/generate_completion.py
INFO:root:etc/crab-bash-completion.sh

Error: $CMSSW_VERSION is not defined. Make sure you do cmsenv first. Exiting...
LapSB:CRABClient$ 
```
As you were away, I forgot to ask for help and/or mention this.

I can fix this with a commit from lxplus, But...
1) it means that there's no hope to do it in `crab_build`
2) even including it in a future automatic pipeline for CRABClient (which Wa has in his plans) may be complicated

belforte:
Done. Thanks @mapellidario for reminding me !

mapellidario (approved):
Looks perfect to me, thanks Stefano! 

Ah yes, you are right, having a cmssw env during the build of the spec files is prohibitive, so much so that you just convinced me that we will never do it :)

PR #5329, opened 2024-08-23
Dryrun as preparelocal

novicecpp:


src/python/CRABClient/ClientUtilities.py
```diff
@@ -788,7 +788,7 @@ def checkStatusLoop(logger, server, api, taskname, targetstatus, cmdname):
     logger.debug("Ended %s process." % ("resubmission" if cmdname == "resubmit" else "submission"))
 
 
-def execute_command(command=None, logger=None, timeout=None, redirect=True):
+def execute_command(command=None, logger=None, timeout=None, redirect=True, live=False):
```
belforte, line 791: what does live do "in English" ?
can you help me remember what "redirect" does ?
where did you plan to use "live" ?

src/python/CRABClient/Commands/preparelocal.py
```diff
@@ -65,10 +68,19 @@ def getInputFiles(self):
         crabDBInfo, _, _ = server.get(api='task', data={'subresource': 'search', 'workflow': taskname})
         status = getColumn(crabDBInfo, 'tm_task_status')
         self.destination = getColumn(crabDBInfo, 'tm_asyncdest')
+        username = getColumn(crabDBInfo, 'tm_username')
+        sandboxName = getColumn(crabDBInfo, 'tm_user_sandbox')
 
         inputsFilename = os.path.join(os.getcwd(), 'InputFiles.tar.gz')
+        sandboxFilename = os.path.join(os.getcwd(), 'sandbox.tar.gz')
         if status == 'UPLOADED':
```
belforte, line 76: do we still need this "if" ? We could always pull files from S3, instead of scheduler.

src/python/CRABClient/Commands/recover.py
belforte, line 1: why is this file changed ? DId you do something, or simply old version from the time you created your branch ?

src/python/CRABClient/ClientUtilities.py
novicecpp: Ah...my bad.
It should be `livelog`: print log as soon as it arrive from subcommand (I plan to use logger instead of print).
`redirect` is you want to capture subprocess stdout/stderr into python variable (available in `proc.communicate()` line below).

src/python/CRABClient/Commands/preparelocal.py
novicecpp: Good catch. 
It should be `if status in ['UPLOADED', 'SUBMITTED']:`.

src/python/CRABClient/Commands/recover.py
novicecpp: Yeah, the old version. No change on this file.

src/python/CRABClient/ClientUtilities.py
belforte: after some testing we think that we do not need `live` option and can simply set `redirect=False`

src/python/CRABClient/Commands/preparelocal.py
belforte: OK

src/python/CRABClient/Commands/recover.py
belforte: OK I will revert

belforte:
implemented slightly differently via #5348

PR #5336, opened 2024-09-30
Adding WAITING status in checkstatusLoop

aspiringmind-code:


src/python/CRABClient/ClientUtilities.py
```diff
@@ -760,10 +760,13 @@ def checkStatusLoop(logger, server, api, taskname, targetstatus, cmdname):
         elif taskStatus in ['NEW', 'HOLDING', 'QUEUED', 'RESUBMIT']:
             logger.info("Please wait...")
             time.sleep(30)
+        elif taskStatus in ['WAITING']:
```
belforte, line 763: I think that this message does not add value here, better simply to add WAITING to the list at line 760
A different story is to write something useful from `crab status`. Also we will remove all of this as we refactor the way to do dry run as per https://github.com/dmwm/CRABServer/issues/6544

belforte (changes requested):

belforte:
you can ignore pylint in CRABClient. Most  of the complain are due to the fact that we need to keep code compatible with python2

belforte (approved):

belforte:
perfect, thanks.

src/python/CRABClient/ClientUtilities.py
aspiringmind-code: Ok, separate elif clause for WAITING removed

PR #5347, opened 2024-11-06
make --dryrun work again with TW v3.2411xx

belforte:
also some long due cleanup


belforte:
need to merge to test with Jenkins

PR #5350, opened 2024-11-21
Python3.12 support: Use importlib instead of deprecated imp module

smuzaffar:
`imp` module has been deprecated and removed in `python 3.12`. This PR proposes to uses `importlib` for `python 3.12` e.g in `CMSSW_14_2_PY3_12_X` IBs. It should not break crab client for CMSSW release with `python version < 3.12`.

This requires newer WMCore specially change in [WMCore/Configuration.py](https://github.com/dmwm/WMCore/pull/11530/files#diff-ae418221154baa0ba96bdcf0566fde2fccf43a95d289f435f28c45e2a0552f1b) . 

Note that using this I was able to submit crab job for CMSSW_14_2_PY312_X but they [fail on crab server side](https://cmsweb.cern.ch:8443/scheddmon/0195/cmsbot/241121_102531:cmsbot_crab_Jenkins_CMSSW_14_2_PY312_X_2024-11-20-2300_el8_amd64_gcc12_123400/job_out.1.0.txt) as due to usage of `imp` in crab server code. So in order to fully support python 3.12 crab jobs, one needs to update crab server code too

PR #5352, opened 2024-11-25
Typo fixed: import load_source from CRABClient

smuzaffar:
This fixes a typo introduced in https://github.com/dmwm/CRABClient/pull/5350


belforte (approved):
Thanks !

PR #5356, opened 2024-11-27
recover should not report FAILED when there is nothing to do. Fix #5355

### LINKED ISSUE #5355: crab recover should not fail if issued before task completed
avoid spurious errors in CI if task execution is delayed.
In this case `crab recover` issues a `kill` but does not wait for e.g. transfers to stop.

belforte:


belforte:
looks very safe to me. Will test in the pipeline once merged.

PR #5362, opened 2024-12-06
add support for SCRAM_MIN_SUPPORTED_MICROARCH. Fix #5354

### LINKED ISSUE #5354: CMSSW_15_0 and above: Support for CMSSW x86-64-v3 as default micro-arch
Hi,

 We are planning to make `x86-64-v3` as default micro-arch for CMSSW_15_0_X ( see https://github.com/dmwm/WMCore/issues/12168 and [core sw meeting](https://indico.cern.ch/event/1482012/)). Though CMSSW_15_0 will support dual micro-arch stack (`x86-64-v3` and `x86-64-v2`) but `x86-64-v3` will be default. 

  In order to avoid increasing the build time (and also crab sandbox size) in cmssw developer area, scram will only build for default micro-arch ( i.e. `x86-64-v3` ). If someone wants to use a `x86-64-v2` resource (e.g. the dataset they need is only available on site with v2 resources) then they have to explicitly enable dual stack build in their developer area.

  We would suggest that, for CMSSW_15_0_X , crab sumbit jobs to `x86-64-v3` resources by default. SCRAM can set some env variable to expose the available micro-archs support e.g for CMSSW_15_0_X and above

- Normal user dev area: SCRAM can set `CMSSW_MICRO_ARCHS=x86-64-v3` i.e. job should run only on `x86-64-v3`
- Dev area multi-archs enabled:  SCRAM can set `CMSSW_MICRO_ARCHS=x86-64-v3,x86-64-v2` i.e job can use both `v2 and v3` resources
- Dev area with multi-archs enabled but user explicitly want to use `x86-64-v2`: SCRAM can set `CMSSW_MICRO_ARCHS=x86-64-v2`

If `CMSSW_MICRO_ARCHS` env sounds reasonable then we can enable this for 15.0.X IBs but if you have any better solutions which can help crab using the correct resources then please do let me know.

belforte:


belforte:
tested with new CRABServer in https://github.com/dmwm/CRABServer/pull/8860

PR #5373, opened 2025-02-17
deprecate numcores option

### LINKED ISSUE #5369: deprecate `--numcores` option for crab resubmit
1) it does not work . [ref](https://github.com/dmwm/CRABServer/issues/8910#issuecomment-2631061776)
2) it makes no sense. We enforce that numCores is equal to numThreads in PSet and PSet can't be changed at resubmit time.

It is not a matter of "giving a few jobs a kick" like with `--maxmemory` or `--maxjobruntime`

aspiringmind-code:
Fix #5369


src/python/CRABClient/ClientMapping.py
belforte, line 1: I think that ClientMapping should not be changed. Did you test this with a config. file which specifies JobType.numCores ? (like CCV test suite)

aspiringmind-code: Oh sorry I thought we were getting rid of the option altogether.

aspiringmind-code: I can submit now specifying the option. Thanks for the review. Merging now.

belforte: it is an interesting questions: why do we allow to specify independently numCores in crabConfig and numThreads in CMSSW's PSet and then we enforce that those are the same ?
I can think of reasons, and maybe one of them was what drove this choice at the time, or maybe it was simply "history".
Generally speaking it makes sense to use CRAB for running non-CMSSW applications with more than one core, but I think that some PSet file is still needed in that case.

PR #5377, opened 2025-03-03
put scriptExe in /debug folder. FIx #5376

### LINKED ISSUE #5376: scriptExe is not available via crab server UI any more
The UI looks in `WEB_DIR/debug` folder, but the script is not in there anymore.
Best solution is to change line 234 in
https://github.com/dmwm/CRABClient/blob/bf347bd46487754cc4388af7a8c77a835496df44/src/python/CRABClient/JobType/UserTarball.py#L219-L236

adding `/debug/` in front of the file name like done for PSet in lines above

belforte:


aspiringmind-code (commented):
@belforte In the issue created, you mentioned making this change to line 234 in addMonFiles(). The current PR makes it beforehand to line 183 in addFiles() itself. I hope addFiles() is called before addMonFiles() to make this ok.

aspiringmind-code:
For example, in [Recover.py](https://github.com/dmwm/CRABClient/blob/bf347bd46487754cc4388af7a8c77a835496df44/src/python/CRABClient/JobType/Recover.py#L158) script, addMonFiles() function is called without addFiles() being called first. Your fix works for Analysis.py script as addFiles() is called first there. Please make sure that Recover.py doesn't give a problem...otherwise LGTM.

belforte:
hmm.. :thinking: 
Thanks for pointing out, I better review. Since I did not test, my bad.

belforte:
Indeed I had initially made the change in the wrong line ! :sob: That's fixed in last commit.
Now I will review what happens in recover.py
Thanks a lot Vijay !

aspiringmind-code (approved):
LGTM. Thanks!

PR #5382, opened 2025-04-01
make crab --version work also w/o cmsenv. Fix #5380

### LINKED ISSUE #5380: crab --version requires cmssw env
I noticed that even running `crab --version` requires cmssw env and it also tries to create `crab.log`  ( which means one can not run it directly from the release area on cvmfs). Is it possible to not depend of cmssw env and create log for ` crab --version` ?

belforte:
still requires $CMSSW_VERSION to be defined

### DIFF
--- a/bin/crab
+++ b/bin/crab
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 script_dir=`dirname $0`
-python_cmd="python"
+python_cmd="python2"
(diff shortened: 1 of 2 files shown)


belforte:
Shahaz says that this is not needed. See in https://github.com/dmwm/CRABClient/issues/5380

PR #5408, opened 2025-08-04
add file size in human units. Fix #5407

### LINKED ISSUE #5407: add file size in human units to checkfile
from
```text
List of disk replicas. Check that file exists and has correct size (2812646607 bytes):
```
to
```text
List of disk replicas. Check that file exists and has correct size (2812646607 bytes) (  2.6 GB):
```

belforte:


aspiringmind-code (approved):
LGTM!

PR #5410, opened 2025-08-11
Make compatible with new tarballs 5409

belforte:


belforte:
@aspiringmind-code I know that this is a tough one to review since context is not clear. All I ask is that you agree that current behavior is preserved. I tested on a task submitted with current TW .. but a second look is always better. If you find that it takes too much of your precious time, do not do it. We will have to go through the whole "new tarballs" thing at some point

aspiringmind-code:
Thanks @belforte I will review it by today.

belforte:
no urgency. I am still way off from having a TW which needs this

aspiringmind-code (approved):
LGTM! Yesterday's walkthrough on the CRABServer side made it much clearer what you are trying to do on the Client side. I have just a minor question on why the need was felt to add the sh interpreter instead of relying on the shebang line inside CMSRunAnalysis.sh

belforte:
Thanks for you review VIjay
about the `sh`... pretty stupid thing. Somehow the file does not have the `+x` permission anymore. I did not have the stomach to look for where it got lost, and since already `gWMS-CMSRunAnalysis.sh` calls this as `sh ./CMSRunAnalysis.sh "$@"` I thought better to simply use the same when running it locally, to be safe.

belforte:
let me know if you think I should change

aspiringmind-code:
No need to change. Thanks!

PR #5418, opened 2025-09-26
py2 compatible HTTPStatus imported.

### LINKED ISSUE #5416: changed for http2 in #5412 do not work on SL6
When using SL6 based (yet still supported) CMSSW releases like CMSSW_10_1 and earlier, CRABClient fails to import `http`
```
  File "/cvmfs/cms-ib.cern.ch/sw/x86_64/nweek-02908/share/cms/crab-dev/v3.250908.00/lib/CRABClient/RestInterfaces.py", line 21, in <module>
    import http
ImportError: No module named http
```

@Lorenzov1999   Please provide a backward-compatible implementation

Until then we need to roll back your #5412  since our intergration tests are failing

nausikt:
Resolve #5416 

I've tested on all slc6 env (e.g. `CMSSW_10_1_0`, `CMSSW_9_4_21`, `CMSSW_8_0_36`). Status responses are fine for every status codes, also in case of unknown one.

### DIFF
--- a/src/python/CRABClient/RestInterfaces.py
+++ b/src/python/CRABClient/RestInterfaces.py
@@ -16,7 +16,8 @@ import logging
 try:
     # Python 2.X
     from urllib import quote as urllibQuote
-    import httplib as _http_client  # pylint: disable=import-error
+    import httplib as _http_client
+    # add an HTTPStatus class such that it can be used like http which is only available in python3
     class HTTPStatus(int):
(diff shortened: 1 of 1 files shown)


belforte:
WOW

please add a comment line of the sort "add an HTTPS method so that it can be used like http which is only available in python3" ... just something to explain what's going on with this new class.

belforte (changes requested):
see comment

PR #5425, opened 2026-03-04
[PEP389] introduce argparse deprecate optparse

### LINKED ISSUE #5424: replace optprase with argparse [PEP389]
Replace **optparse** with **argparse** for long-term support, with Python 2.7 as the minimum supported version (PEP 389).

See example on CRABServer side, https://github.com/dmwm/CRABServer/pull/9261

nausikt:
Resolve https://github.com/dmwm/CRABClient/issues/5424

The hardest part is preserving the strict `disable_interspersed_args` behaviour and adding back the validating of unknown flags in [CRABOptParser.py](https://github.com/dmwm/CRABClient/pull/5425/changes#diff-7f3c43626918e298fd25bab96b58c98477c511f0d37dffb45064238daa34484c). 
*Thanks to LLM, helping me polish this part meanwhile we both looking at original [optparse.py](https://github.com/python/cpython/blob/dc12d1999b88e84d5a6b8e491be468b73379e54b/Lib/optparse.py#L1395C5-L1445C1)*

So, CRABOptParser in general + submit, createmyproxy subcommands will behave exactly the same.
*See also [the important bit disable_interspersed_args](https://docs.python.org/3/library/optparse.html#optparse.OptionParser.disable_interspersed_args)*

*P.S. Let me exhaustively test first (may write some unit-tests), do we already have test cases covered these behaviors?*
*P.S. Besides that, I think my proposed abstraction/structure is fairly minimal and clean. what do you think?*

### DIFF
--- a/bin/crab.py
+++ b/bin/crab.py
@@ -60,7 +60,7 @@ class CRABClient(object):
 
     def __call__(self):
 
-        (options, args) = self.parser.parse_args()
+        (options, args) = self.parser.parse_cmd()
 
         ## The default logfile destination is ./crab.log. It will be changed once we
         ## know/create the CRAB project directory.
--- a/doc/generate_modules.py
+++ b/doc/generate_modules.py
@@ -232,7 +232,7 @@ def main():
     """
     Parse and check the command line arguments.
     """
-    parser = argparse.ArgumentParser(usage="""usage: %prog [options] <package path> [exclude paths, ...]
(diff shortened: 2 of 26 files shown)


belforte:
Hi @nausikt I tried to understand this `disable_interspersed_args` but it requires too much time. I think I have looked at this in the pasta and it may be that the only reason for that is that original developers wanted to have
`crab --debug <command> <other options>`
And I have no idea if anybody is actually using that, other than us in some test scripts.
If all that we need to simplify our life is to move that to be
`crab command --debug --this --that`
I am surely in favor !

PR #5428, opened 2026-06-28
add some comments

belforte:


belforte:
trivial PR, but I need a review !

nausikt (approved):
Thanks for the comments!

nausikt:
Another workaround might be to bypass this by using a maintainer role.

belforte:
hmm.. would that also bypass the reason why we put in this restriction in the first place ? IIUC the concern is that someone steals my GH identity and push malicious code in. As it happened to Valentin with the [Megalodon](https://www.ossprey.com/blog/megalodon-credential-harvester-attack) attack

PR #5429, opened 2026-06-29
Fix #5427 handle case of incorrect file counting from FJR

belforte:
Fixes #5427 
incorrect file (and event) counting happens when FJR contains wrong information due to bad input file naming or is simply faked by users.
If that happens, warn the user and avoid printing misleading information in `crab report`


belforte:
Need to verify compatibility with older python and adapt if needed

belforte:
tested with both python3 and python2.7

nausikt (approved):
Looks good to me!

